PR #1, opened 2015-02-20
Initial implementation of CORS

Praburaj:
This includes a CORS middleware
CORS related libraries for building and validating policies
A bunch of tests


Praburaj:
@harshgMSFT @yishaigalatzer @muratg @lodejard @Eilon

src/Microsoft.AspNet.Cors/CorsMiddleware.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            _options.PolicyProvider = _options.PolicyProvider ?? new CorsPolicyProvider();
+            _options.CorsEngine = _options.CorsEngine ?? new CorsEngine();
+        }
+
+        public async Task Invoke(HttpContext context)
+        {
+            var corsRequestContext = new CorsRequestContext(context);
+
+            if (corsRequestContext.IsCorsRequest)
```
troydai: Create a new instance of context to check if a request is a Cors request looks like an overkill, since it only takes reading a header. We can make the `IsCorsRequest` a static method. If the request is indeed a Cors request then create the `CorsRequestContext`. It avoid unnecessary allocation.

src/Microsoft.AspNet.Cors/CorsMiddleware.cs
```diff
@@ -0,0 +1,92 @@
...
+                    {
+                        // Always respond to preflight requests.
+                        if (corsResult.IsValid)
+                        {
+                            context.Response.StatusCode = StatusCodes.Status200OK;
+                            WriteCorsHeaders(context, corsResult);
+                        }
+                        else
+                        {
+                            context.Response.StatusCode = StatusCodes.Status400BadRequest;
```
troydai: The response may not necessarily always be 400. It may be the way in which we process incorrect preflight request but after read the spec for another time I think it's debatable.
http://www.w3.org/TR/cors/#resource-preflight-requests

The words used in the spec is:
> The request is outside the scope of this specification.

I tend to interrupt it as: __treat it as a normal request come in the pipeline and see how user logic deal with it__. So even we don't process it again, the return status code should be 405 method not allowed as user tends not to define how to process OPTION.

This is just my suggestion. Practically, returns 400 may not cause any trouble, because for any status out of 2XX range user agent will simply stop the processing.

src/Microsoft.AspNet.Cors/CorsOptions.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        /// <summary>
+        /// The CORS engine.
+        /// </summary>
+        public ICorsEngine CorsEngine { get; set; }
+
+        /// <summary>
+        /// A policy that allows all headers, all methods, any origin and supports credentials
+        /// </summary>
+        public static CorsOptions AllowAll
```
troydai: Security review warning. We taken out setting like this and force user to explicit call out accept all in previous design after security review. This option is not as bad as accept all by default but is still way too easy to use and I'm afraid user will easily fall in love with the always allow all option.

blowdart: You can just @ me and I appear. It was luck this was at the top of my inbox.

So yea, remove that property. This is one thing we don't want to make easy.

src/Microsoft.AspNet.Cors/CorsMiddleware.cs
Praburaj: I can fix that. This is all about checking for presence of origin header.

src/Microsoft.AspNet.Cors/CorsOptions.cs
Praburaj: @troydai @blowdart thanks for jumping in :)

Sure. I can remove that. Katana had this option - so I did a similar thing.

blowdart:
@Praburaj Heh, we can correct Katana's mistakes :)

samples/CORSClient/Startup.cs
```diff
@@ -0,0 +1,41 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNet.Builder;
+using Microsoft.AspNet.Diagnostics;
+using Microsoft.Framework.DependencyInjection;
+using Microsoft.Framework.Logging;
+using Microsoft.Framework.Logging.Console;
+
+namespace Client
```
troydai: Give it a more sophisticated namespace even it is a sample.

samples/CORSClient/project.json
```diff
@@ -0,0 +1,29 @@
...
+    "frameworks": {
+        "aspnet50": {},
+        "aspnetcore50": {}
+    },
+    "exclude": [
+        "wwwroot"
+    ],
+    "bundleExclude": [
+        "node_modules",
+        "bower_components",
```
troydai: We need these two?

samples/CORSClient/Startup.cs
Praburaj: Yeah. It will be CORSClient. I renamed the project missed this.

samples/CORSClient/project.json
Praburaj: I can remove those. Was out of the default templates.

src/Microsoft.AspNet.Cors.Core/CorsPolicy.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// <inheritdoc />
+        public IList<string> Headers { get; private set; } = new List<string>();
+
+        /// <inheritdoc />
+        public IList<string> Methods { get; private set; } = new List<string>();
+
+        /// <inheritdoc />
+        public IList<string> Origins { get; private set; } = new List<string>();
+
+        private long? _preflightMaxAge;
```
troydai: private field should be placed at beginning.

src/Microsoft.AspNet.Cors.Core/CorsPolicyProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        /// <summary>
+        /// A pluggable callback that will be used to select the CORS policy for the given requests.
+        /// </summary>
+        public Func<ICorsRequestContext, Task<ICorsPolicy>> PolicyResolver { get; set; }
+
+        /// <inheritdoc />
+        public virtual Task<ICorsPolicy> GetCorsPolicyAsync(ICorsRequestContext context)
+        {
+            return PolicyResolver(context);
```
troydai: Through the `PolicyResovler` is assigned, it sill worth to guard for `null`

src/Microsoft.AspNet.Cors.Core/CorsRequestContext.cs
```diff
@@ -0,0 +1,68 @@
...
+                RequestPath = context.Request.Path;
+                HttpMethod = context.Request.Method;
+                Host = context.Request.Host.Value;
+                AccessControlRequestMethod = context.Request.Headers.Get(CorsConstants.AccessControlRequestMethod);
+                AccessControlRequestHeaders = context.Request.Headers.GetCommaSeparatedValues(CorsConstants.AccessControlRequestHeaders);
+            }
+        }
+
+        /// <inheritdoc />
+        public IList<string> AccessControlRequestHeaders { get; set; }
```
troydai: Need to the public settable?

src/Microsoft.AspNet.Cors/CorsMiddleware.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// <param name="options"></param>
+        public CorsMiddleware(
+            [NotNull]RequestDelegate next,
+            [NotNull]CorsOptions options)
+        {
+            _next = next;
+            _options = options;
+
+            _options.PolicyProvider = _options.PolicyProvider ?? new CorsPolicyProvider();
+            _options.CorsEngine = _options.CorsEngine ?? new CorsEngine();
```
troydai: Why not just make these two injected in DI?

src/Microsoft.AspNet.Cors/CorsOptions.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        /// <summary>
+        /// The CORS policy to apply.
+        /// </summary>
+        public ICorsPolicyProvider PolicyProvider { get; set; }
+
+        /// <summary>
+        /// The CORS engine.
+        /// </summary>
+        public ICorsEngine CorsEngine { get; set; }
```
harshgMSFT: Not services lets remove this an enable DI.

src/Microsoft.AspNet.Cors/CorsOptions.cs
Praburaj: Removed.

src/Microsoft.AspNet.Cors.Core/CorsPolicyProvider.cs
Praburaj: I would say let it throw so that we know its not configured.

src/Microsoft.AspNet.Cors.Core/CorsRequestContext.cs
Praburaj: I think its no harm. Its useful for testing :)

src/Microsoft.AspNet.Cors/CorsMiddleware.cs
Praburaj: When the preflight request is handled by MVC (in future) they will let the request flow to the next middleware in the pipeline if the action finder cannot find a matching action as requested by preflight request. But on middleware where there is just one policy it will handle it always.

Praburaj:
Updated PR.

harshgMSFT:
Closing this one as will be sending a separate PR with just the common changes.

PR #2, opened 2015-02-24
Initial Commit to enable CI

harshgMSFT:


src/Microsoft.AspNet.Cors.Core/project.json
```diff
@@ -0,0 +1,17 @@
+﻿{
+	"version": "1.0.0-*",
```
Eilon: Tabs :frowning: :panda_face:

harshgMSFT: my bad, it showed up fine in my vs ( looks like for some reason tabs vs spaces was messed up). Ill fix it along with the next change.

PR #3, opened 2015-02-25
Adding Base interfaces and implementation for Cors.Core

harshgMSFT:
Also adds tests for the same.

cc: @Eilon , @lodejard , @davidfowl , @Praburaj


global.json
```diff
@@ -1,3 +1,7 @@
 {
-    "sources": ["src"]
+	"sources": [
+		"src",
+		"samples",
```
harshgMSFT: no samples yet remove samples.

src/Microsoft.AspNet.Cors.Core/CorsResult.cs
```diff
@@ -0,0 +1,179 @@
...
+        }
+
+        /// <summary>
+        /// Returns CORS-specific headers that should be added to the response.
+        /// </summary>
+        /// <returns>The response headers.</returns>
+        public virtual IDictionary<string, string> ToResponseHeaders()
+        {
+            var headers = new Dictionary<string, string>();
+            if (!IsValid)
```
harshgMSFT: This is a change in behavior from 5.0 , and this one is according to cors spec.

test/Microsoft.AspNet.Cors.Core.Test/project.json
```diff
@@ -0,0 +1,21 @@
+﻿{
+    "version": "1.0.0-*",
+
+	"dependencies": {
```
harshgMSFT: fix it.

src/Microsoft.AspNet.Cors.Core/CorsPolicyBuilder.cs
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+        /// <summary>
+        /// Adds the specified <paramref name="origins"/> to the policy.
+        /// </summary>
+        /// <param name="origins">The origins that are allowed.</param>
+        /// <returns>The current policy builder</returns>
+        public CorsPolicyBuilder AddOrigins(params string[] origins)
+        {
+            foreach (var req in origins)
```
kichalla: nit: AddRange

src/Microsoft.AspNet.Cors.Core/EnableCorsAttribute.cs
```diff
@@ -0,0 +1,25 @@
...
+        /// <summary>
+        /// Creates a new instance of the <see cref="EnableCorsAttribute"/>.
+        /// </summary>
+        /// <param name="policyName">The name of the policy to be applied.</param>
+        public EnableCorsAttribute(string policyName)
+        {
+            PolicyName = policyName;
+        }
+
+
```
harshgMSFT: extra line.

src/Microsoft.AspNet.Cors.Core/CorsOptions.cs
```diff
@@ -0,0 +1,50 @@
...
+using Microsoft.Framework.Internal;
+
+namespace Microsoft.AspNet.Cors.Core
+{
+    /// <summary>
+    /// Provides programmatic configuration for Cors.
+    /// </summary>
+    public class CorsOptions
+    {
+        private IDictionary<string, CorsPolicy> PolicyMap { get; } = new Dictionary<string, CorsPolicy>();
```
kichalla: policy name - case insensitive?

src/Microsoft.AspNet.Cors.Core/EnableCorsAttribute.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNet.Cors.Core
+{
+    /// <inheritdoc />
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = false)]
```
kichalla: inherited = true? (scenario: base controller could have a cors policy which applies to derived controllers)

harshgMSFT: isnt that on by default?

kichalla: you mean like the call to GetCustomAttributes(inherit: true) will get it?...i don not think so....probably a functional test in MVC could be useful here... 

I get confused with inherited, but looks like i am not the only one...take a look at the matrix :smile: 
http://blogs.msdn.com/b/jmstall/archive/2013/07/17/attributeusage-inherited-flag.aspx

harshgMSFT: I am sorry my bad, i thought i had not marked it as inherited false, (hence the comment about default).. yes you are right, this will be fixed.

src/Microsoft.AspNet.Cors.Core/CorsPolicyBuilder.cs
```diff
@@ -0,0 +1,193 @@
...
+            _policy.Origins.Clear();
+            _policy.Origins.Add(CorsConstants.AnyOrigin);
+            return this;
+        }
+
+        /// <summary>
+        /// Ensures that the policy allows any method.
+        /// </summary>
+        /// <returns>The current policy builder</returns>
+        public CorsPolicyBuilder AllowAnyMehtod()
```
borgdylan: AllowAnyMethod?

src/Microsoft.AspNet.Cors.Core/CorsEngine.cs
```diff
@@ -0,0 +1,187 @@
...
+using Microsoft.AspNet.Http;
+using Microsoft.Framework.Internal;
+using Microsoft.Framework.OptionsModel;
+
+namespace Microsoft.AspNet.Cors.Core
+{
+    /// <summary>
+    /// Default implementation of <see cref="ICorsEngine"/>.
+    /// </summary>
+    public class CorsEngine : ICorsEngine
```
lodejard: rename `CorsService : ICorsService` to match pattern like `AuthorizationService : IAuthorizationService`

global.json
davidfowl: Is this using tabs?

src/Microsoft.AspNet.Cors.Core/CorsResult.cs
```diff
@@ -0,0 +1,179 @@
...
+
+        /// <summary>
+        /// Gets the allowed headers that can be exposed on the response.
+        /// </summary>
+        public IList<string> AllowedExposedHeaders { get; } = new List<string>();
+
+        /// <summary>
+        /// Gets or sets the number of seconds the results of a preflight request can be cached.
+        /// </summary>
+        public long? PreflightMaxAge
```
lodejard: Would `TimeSpan?` instead of `'long?` make sense, to make the units explicit?

src/Microsoft.AspNet.Cors.Core/CorsEngine.cs
```diff
@@ -0,0 +1,187 @@
...
+            [NotNull] CorsResult result)
+        {
+            var origin = context.Request.Headers.Get(CorsConstants.Origin);
+            if (origin != null)
+            {
+                if (policy.AllowAnyOrigin)
+                {
+                    if (policy.SupportsCredentials)
+                    {
+                        result.AllowedOrigin = origin;
```
lodejard: Should we also include `Vary: Origin` in the result?

http://www.w3.org/TR/cors/#resource-implementation

src/Microsoft.AspNet.Cors.Core/RequestExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// </summary>
+    public static class RequestExtensions
+    {
+        /// <summary>
+        /// Determines if the current <see cref="HttpRequest"/> is a pre-flight request.
+        /// </summary>
+        /// <param name="request">The given <see cref="HttpRequest"/>.</param>
+        /// <returns><c>true</c> if there is an Origin header and there is a non null Access-Control-Request-Method
+        /// header and the request method is OPTIONS.<c>false</c> otherwise.</returns>
+        public static bool IsPreflight(this HttpRequest request)
```
lodejard: Don't use extension methods for this, just look for method == "OPTIONS" at the appropriate place.

src/Microsoft.AspNet.Cors.Core/RequestExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+                accessControlRequestMethod != null &&
+                string.Equals(request.Method, CorsConstants.PreflightHttpMethod, StringComparison.Ordinal);
+        }
+
+        /// <summary>
+        /// Determines if the current <see cref="HttpRequest"/> is a CORS request.
+        /// </summary>
+        /// <param name="request">The given <see cref="HttpRequest"/></param>
+        /// <returns><c>true</c> if the request contains Origin header.</returns>
+        public static bool IsCorsRequest(this HttpRequest request)
```
lodejard: Don't expose extension method for this either.

src/Microsoft.AspNet.Cors.Core/CorsResult.cs
```diff
@@ -0,0 +1,179 @@
...
+                }
+                _preflightMaxAge = value;
+            }
+        }
+
+        /// <summary>
+        /// Returns CORS-specific headers that should be added to the response.
+        /// </summary>
+        /// <returns>The response headers.</returns>
+        public virtual IDictionary<string, string> ToResponseHeaders()
```
lodejard: Move this logic to `ICorsEngine.ApplyResult(HttpResponse, CorsResult)` method - data objects should avoid having behavior.

src/Microsoft.AspNet.Cors.Core/CorsEngine.cs
```diff
@@ -0,0 +1,187 @@
...
+        /// used by the caller to set apporpriate response headers.</returns>
+        public CorsResult EvaluatePolicy([NotNull] HttpContext context, string policyName)
+        {
+            var policy = _options.GetPolicy(policyName);
+            return EvaluatePolicy(context, policy);
+        }
+
+        /// <inheritdoc />
+        public CorsResult EvaluatePolicy([NotNull] HttpContext context, [NotNull]CorsPolicy policy)
+        {
```
lodejard: This should be changed to call two different methods, `EvaluateRequest(context, policy, result)` and `EvaluatePreflightRequest(context, policy, result)`. The `EvaluateRequest` should follow the steps defined by http://www.w3.org/TR/cors/#resource-requests and `EvaluatePreflightRequest` should follow http://www.w3.org/TR/cors/#resource-preflight-requests.

EvaluatePolicy calls EvaluatePreflightRequest when http method is `OPTIONS` and the request headers contain `Access-Control-Request-Method`, otherwise EvaluateRequest.

src/Microsoft.AspNet.Cors.Core/CorsEngine.cs
```diff
@@ -0,0 +1,187 @@
...
+
+            if (!TryValidateOrigin(context, policy, corsResult))
+            {
+                return corsResult;
+            }
+
+            corsResult.SupportsCredentials = policy.SupportsCredentials;
+            if (context.Request.IsPreflight())
+            {
+                if (!TryValidateMethod(context, policy, corsResult))
```
lodejard: When implementing EvaluatePreflightRequest following http://www.w3.org/TR/cors/#resource-preflight-requests be sure that steps 1-6 are all followed before any of the result properties are assigned.

src/Microsoft.AspNet.Cors.Core/CorsResult.cs
```diff
@@ -0,0 +1,179 @@
...
+            get
+            {
+                return ErrorMessages.Count == 0;
+            }
+        }
+
+        /// <summary>
+        /// Gets the error messages.
+        /// </summary>
+        public IList<string> ErrorMessages { get; } = new List<string>();
```
lodejard: We should remove ErrorMessages. The only two outcomes of the CORS algorithm should be adding additional headers to a simple request (simple requests are never prevented) and a 204 response for a preflight request. An OPTIONS returning 204 without Access-Control-Allow-Origin should be interpreted as a preflight request failure by the caller.

src/Microsoft.AspNet.Cors.Core/CorsPolicyBuilder.cs
harshgMSFT: yeah.. fixed..

src/Microsoft.AspNet.Cors.Core/CorsEngine.cs
harshgMSFT: done.

src/Microsoft.AspNet.Cors.Core/CorsEngine.cs
harshgMSFT: Good catch yes we should. Added.

src/Microsoft.AspNet.Cors.Core/CorsResult.cs
harshgMSFT: ```(simple requests are never prevented)```
I thought we prevented simple requests (and non simple non preflight requests) which do not match the policy. i.e set the response code as 400 without any additional methods.
If we do not stop execution of simple requests then a POST with form data can be executed.

src/Microsoft.AspNet.Cors.Core/CorsEngine.cs
harshgMSFT: done

src/Microsoft.AspNet.Cors.Core/EnableCorsAttribute.cs
harshgMSFT: fixed.

src/Microsoft.AspNet.Cors.Core/CorsPolicyBuilder.cs
harshgMSFT: doesnt work on IList it has to be a list for add range to work.

src/Microsoft.AspNet.Cors.Core/CorsResult.cs
harshgMSFT: sure ...

src/Microsoft.AspNet.Cors.Core/CorsOptions.cs
harshgMSFT: I think it should be case sensitive. Its same as authorization policy.

harshgMSFT:
@lodejard updated.

src/Microsoft.AspNet.Cors.Core/CorsOptions.cs
rynowak: In general programmatic names are case-sensitive

src/Microsoft.AspNet.Cors.Core/CorsPolicyBuilder.cs
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        /// <summary>
+        /// Ensures that the policy allows any origin.
+        /// </summary>
+        /// <returns>The current policy builder</returns>
+        public CorsPolicyBuilder AllowAnyOrigin()
+        {
+            _policy.Origins.Clear();
+            _policy.Origins.Add(CorsConstants.AnyOrigin);
```
rynowak: Any concern for the case where a user adds `foo.com`, `bar.edu`, and `*` using AddOrigins? Looks like `CorsPolicy.AllowAnyOrigin` will not recognize this, but nothing stops you from configuring it. Similar cases are possible with the other properties that support wildcards.

harshgMSFT: I thought about this, but this did not seem to be a big problem, mainly because the user agent is not expected to send a value * for the origin, header, or method, which means if you have a policy which has ```foo.com```, ```bar.edu```, and ```*```, in the policy, we would treat ```*``` as just another value. Now we might argue and say if a user adds a ```*``` using ```AddOrigins``` should we clear the Origins and allow any origin... to me it seems like too much smartness.

Having said that I was thinking of adding some validation for CorsPolicy which would validate things like an origin of the form ```http://foo.com/bar``` are not valid, in addition to origins having a ```*``` value.

harshgMSFT:
@Eilon ping ...

src/Microsoft.AspNet.Cors.Core/CorsResult.cs
lodejard: Simple POST with form data are prevented because CSRF token is missing, not because CORS processing is comparing policy. With simple requests stopping requests which don't match CORS policy doesn't really prevent attacks. As I understand it, the javascript would just as easily add a script src or create and submit form tag in that case to cause the request to be sent without the Origin header, bypassing the CORS policy anyway. The real feature is that if the javascript attempts any XHR or ajax call to get the data from the server, the browser won't share the results with the calling javascript.

harshgMSFT: Sure .. I agree. The only reason I brought that up was I thought in the meeting we agreed to block the call. I will update the other two PRs ( MVC and the CORS Middleware) to reflect this.

harshgMSFT:
@Eilon @lodejard .. updated. I have addressed all feedback please let me know if there is anything else which you find is missing.

test/Microsoft.AspNet.Cors.Core.Test/CorsServiceTests.cs
```diff
@@ -0,0 +1,904 @@
...
+            
+            // Assert
+            Assert.Empty(result.AllowedHeaders);
+            Assert.Empty(result.AllowedMethods);
+            Assert.Empty(result.AllowedExposedHeaders);
+            Assert.Null(result.AllowedOrigin);
+        }
+
+        [Fact]
+        public void EaluatePolicy_DoesCaseSensitiveComparison()
```
Eilon: Typo in method name.

Eilon:
Looks good to me, I'll try to find @lodejard today to make sure he has time to take a final look.

src/Microsoft.AspNet.Cors.Core/CorsServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// <summary>
+        /// Add services needed to support CORS to the given <paramref name="serviceCollection"/>.
+        /// </summary>
+        /// <param name="serviceCollection">The service collection to which CORS services are added.</param>
+        /// <param name="config">An optional configuration which can be applied to the <see cref="ServiceDescriber"/>.
+        /// </param>
+        /// <returns>The updated <see cref="IServiceCollection"/>.</returns>
+        public static IServiceCollection AddCors(this IServiceCollection serviceCollection, IConfiguration config = null)
+        {
+            var describe = new ServiceDescriber(config);
```
davidfowl: Are you working off old bits? ServiceDescriber is gone.

src/Microsoft.AspNet.Cors.Core/project.json
```diff
@@ -1,7 +1,12 @@
 {
-	"version": "1.0.0-*",
-	"dependencies": {
-	},
+    "version": "1.0.0-*",
+    "dependencies": {
+         "Microsoft.Framework.ConfigurationModel": "1.0.0-*",
+         "Microsoft.Framework.DependencyInjection": "1.0.0-*",
```
davidfowl: You should only depend on Microsoft.Framework.DependencyInjection.Interfaces and Microsoft.Framework.ConfigurationModel.Interfaces

src/Microsoft.AspNet.Cors.Core/CorsServiceCollectionExtensions.cs
dougbu: @davidfowl fyi `ServiceDescriber` hasn't yet left the building.  I should be pushing the MVC PR (#2137) in a bit.

@harshgMSFT this code should just do `serviceCollection.AddTransient<ICorsService, CorsService>();` or perhaps `TryAdd(anotherCollection)` where `var anotherCollection = new ServiceCollection();`.

in addition, you shouldn't need an overload accepting an `IConfiguration` parameter.

lodejard:
:shipit: assuming other threads sorted out

PR #4, opened 2015-02-27
Adding Cors Middleware.

harshgMSFT:
cc: @Eilon ,@loudj, @DamianEdwards,@davidfowl, @Praburaj


samples/UsePolicy/Startup.cs
```diff
@@ -0,0 +1,23 @@
...
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddCors();
+        }
+
+        public void Configure(IApplicationBuilder app)
+        {
+            var allowSingleOrigin = new CorsPolicy();
+            allowSingleOrigin.Origins.Add("http://example.com");
+            app.UseCors(allowSingleOrigin);
```
davidfowl: This needs to be a one liner.

harshgMSFT: well i should have also added the builder pattern based overload as a sample. Here is what it looks like
```c#
app.UseCors(builder => builder.AddOrigins("http://example.com"));
```
does that sound okay?

src/Microsoft.AspNet.Cors.Core/ICorsPolicyProvider.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// A type which can provide a <see cref="CorsPolicy"/> for a particular <see cref="HttpContext"/>.
+    /// </summary>
+    public interface ICorsPolicyProvider
+    {
+        /// <summary>
+        /// Gets a <see cref="CorsPolicy"/> from the given <paramref name="context"/>
+        /// </summary>
+        /// <param name="context">The <see cref="HttpContext"/> associated with this call.</param>
+        /// <returns>A <see cref="CorsPolicy"/></returns>
+        CorsPolicy GetPolicy(HttpContext context);
```
kichalla: While in Web API we came across a scenario where users wanted to get the policy from database (ex: one could dynamically add/update trusted origin list in a database without changing code)...for this reason I *think* we had this call to be async...

samples/UsePolicy/Startup.cs
Eilon: I think @davidfowl has opinions on the correct `.Use...` usage pattern.

src/Microsoft.AspNet.Cors.Core/CorsOptions.cs
```diff
@@ -13,8 +13,30 @@ namespace Microsoft.AspNet.Cors.Core
...
+            {
+                return _defaultPolicyName;
+            }
+
+            set
+            {
+                if (value == null)
+                {
+                    throw new ArgumentNullException("value",
+                                                    Resources.FormatPropertyOfTypeCannotBeNull(
```
Eilon: Don't need a fancy exception message, just throw a plain old ArgNullException with the param name `value` and that's it.

src/Microsoft.AspNet.Cors.Core/Resources.resx
```diff
@@ -117,6 +117,9 @@
   <resheader name="writer">
       <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="PropertyOfTypeCannotBeNull" xml:space="preserve">
```
Eilon: Remove per my other comment.

src/Microsoft.AspNet.Cors.Core/ICorsPolicyProvider.cs
troydai: :+1:

src/Microsoft.AspNet.Cors/CorsMiddleware.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// <param name="next"></param>
+        /// <param name="engine"></param>
+        /// <param name="options"></param>
+        /// <param name="policyName"></param>
+        public CorsMiddleware(
+           [NotNull] RequestDelegate next,
+           [NotNull] ICorsEngine engine,
+           [NotNull] IOptions<CorsOptions> options,
+           [NotNull] string policyName)
+            : this(next,engine, options.Options.GetPolicy(policyName))
```
troydai: add a space before `engine`.

test/Microsoft.AspNet.Cors.Core.Test/CorsOptionsTests.cs
```diff
@@ -0,0 +1,20 @@
...
+        [Fact]
+        public void CookieName_SettingNullValue_Throws()
+        {
+            // Arrange
+            var options = new CorsOptions();
+
+            // Act & Assert
+            var ex = Assert.Throws<ArgumentNullException>(() => options.DefaultPolicyName = null);
+            Assert.Equal("The 'DefaultPolicyName' property of 'Microsoft.AspNet.Cors.Core.CorsOptions' must not be null." +
+                         Environment.NewLine + "Parameter name: value", ex.Message);
```
troydai: Validate the message of an exception doesn't seem to be a stable tests, it may break easily when product is localized , the type name is changed or the format is updated.

test/Microsoft.AspNet.Cors.Test/CorsMiddlewareTests.cs
```diff
@@ -0,0 +1,217 @@
...
+                app.Run(async context =>
+                {
+                    await context.Response.WriteAsync("Cross origin response");
+                });
+            }))
+            {
+                // Act
+                // Preflight request.
+                var response = await server.CreateRequest("/")
+                    .AddHeader(CorsConstants.Origin, "http://localhost:5001/sub")
```
troydai: `http://localhost:5001/sub` can't be an origin. In a real world, no user agent will sent a preflight request like this. An origin is a combination of URI scheme, hostname, and port number. http://www.w3.org/Security/wiki/Same_Origin_Policy

test/Microsoft.AspNet.Cors.Core.Test/CorsOptionsTests.cs
Eilon: We do test exception messages where appropriate. However, in this case as I mentioned earlier this shouldn't have a custom exception message anyway.

test/Microsoft.AspNet.Cors.Test/CorsMiddlewareTests.cs
harshgMSFT: :+1: will fix.

src/Microsoft.AspNet.Cors.Core/CorsOptions.cs
harshgMSFT: I am fine removing it, but we do it at other places today. (and I think i got a comment previously which was the opposite guideline) eg. https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/AntiForgery/AntiForgeryOptions.cs#L41-L44

Eilon: Yeah those comments were incorrect. We shouldn't be using any special messages for such common exceptions.

samples/UsePolicy/Startup.cs
PinpointTownes: To match the authentication middleware style, I think it would be better to take an `Action<CorsOptions>` instead:

```csharp
app.UseCors(options => {
    options.Origins.Add("http://example.com");
});
```

harshgMSFT: Origins etc are defined in the policy and are not options. The pattern is to pass the builder as I pointed above. Ill include that in the sample.

harshgMSFT: we chose to stick with 
```app.UseCors(builder => builder.AddOrigins("http://example.com"));```
and remove 
the one which takes in the policy object directly.

File: src/Microsoft.AspNet.Cors/CorsMiddleware.cs (hunk 5 above)
Comment (harshgMSFT) [inline src/Microsoft.AspNet.Cors/CorsMiddleware.cs]: fixed.

**Comment by harshgMSFT** [comment]
branch updated.

File: src/Microsoft.AspNet.Cors.Core/CorsResult.cs [hunk 8]
```diff
@@ -0,0 +1,179 @@
...
+    /// Results returned by <see cref="ICorsEngine"/>.
+    /// </summary>
+    public class CorsResult
+    {
+        private long? _preflightMaxAge;
+
+        /// <summary>
+        /// Gets a value indicating whether the result is valid.
+        /// </summary>
+        public bool IsValid
```
lodejard: i'm confused about https://github.com/aspnet/CORS/pull/3 and https://github.com/aspnet/CORS/pull/4 - is one of them out of date? do we still have ErrorMessages on the CorsResult?

harshgMSFT:
@Eilon @lodejard updated with the latest changes.

Eilon:
Looks good to me.

harshgMSFT:
@lodejard ping.

CORS.sln
```diff
@@ -10,11 +10,21 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
...
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.AspNet.Cors.Core", "src\Microsoft.AspNet.Cors.Core\Microsoft.AspNet.Cors.Core.xproj", "{C573AEE1-8D54-4A83-8D6B-61C85E8F713E}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "test", "test", "{F32074C7-087C-46CC-A913-422BFD2D6E0A}"
 EndProject
-Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.AspNet.Cors.Core.Test", "test\Microsoft.AspNet.Cors.Core.Test\Microsoft.AspNet.Cors.Core.Test.kproj", "{B4F83A06-EB8E-4186-84C4-C6DAF7EB03D4}"
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.AspNet.Cors.Core.Test", "test\Microsoft.AspNet.Cors.Core.Test\Microsoft.AspNet.Cors.Core.Test.xproj", "{B4F83A06-EB8E-4186-84C4-C6DAF7EB03D4}"
+EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.AspNet.Cors", "src\Microsoft.AspNet.Cors\Microsoft.AspNet.Cors.kproj", "{41349FCD-D1C4-47A6-82D0-D16D00A8D59D}"
+EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.AspNet.Cors.Test", "test\Microsoft.AspNet.Cors.Test\Microsoft.AspNet.Cors.Test.kproj", "{F05BE96F-F869-4408-A480-96935B4835EE}"
```
lodejard: Several projects in the .sln are still kproj though the files have been renamed

samples/UsePolicyBuilder/Startup.cs
```diff
@@ -0,0 +1,21 @@
...
+    public class Startup
+    {
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddCors();
+        }
+
+        public void Configure(IApplicationBuilder app)
+        {
+            app.UseCors(builder => builder.AddOrigins("http://example.com"));
```
lodejard: `policy => policy.xxx` instead of builder would be clearer. (I know, it's just sample :) )

test/Microsoft.AspNet.Cors.Test/CorsMiddlewareTests.cs
```diff
@@ -0,0 +1,214 @@
...
+    {
+        [Fact]
+        public async Task CorsRequest_MatchPolicy_SetsResponseHeaders()
+        {
+            // Arrange
+            using (var server = TestServer.Create(app =>
+            {
+                app.UseServices(services => services.AddCors());
+                app.UseCors(builder => 
+                    builder.AddOrigins("http://localhost:5001")
```
lodejard: I like it!

Do we have other examples of builder (chaining) APIs? I think they are more like `.WithOrigins(...).WithMethods(...)`

samples/UsePolicyBuilder/Startup.cs
harshgMSFT: :+1:

CORS.sln
harshgMSFT: yeah, need to clean up, some problem with tooling will fix before checkin.

test/Microsoft.AspNet.Cors.Test/CorsMiddlewareTests.cs
harshgMSFT: I am just thinking should we just say `.AllowOrigins(...).AllowMethods(...)` etc?

harshgMSFT:
da6a4f2ab7cf75b99391a1f93dcbb161d45109d8

PR #21, opened 2015-05-22
Fix AllowAnyHeader error "SEC7123: Request header *** was not present…

fuxuan:
should use AppendCommaSeparatedValues to add response headers.


Eilon:
@kichalla can you take a look at this as well?

kichalla:
Thanks for your contribution @fuxuan but currently the dev branch has these changes already.

Issue #22, opened 2015-06-10
Access-Control-Allow-Headers shows as two separate headers which makes IE not process a cors response.


harshgMSFT:
Moving the bug from https://github.com/aspnet/Home/issues/640
@esyorcho 
I've been trying to "tie" both headers together (authentication and content-type), so that they show up in one line like:

Access-Control-Allow-Headers: content-type, authorization

But no matter what I did, they were being separated in 2 lines. I removed the authorization header from my first call (the login) since I didn't need it anyways, and then it went through (because it only had the content-type header). Somehow post calls within the application that need the authorization header still have the same problem (content-type missing, because it's in the second line and IE can't read it). Could you please tell me if there's a way that you know to make both headers show up in one line? At the moment we're adding the Authorization header this way in js:

                // add authentication header:
                var authHeader = "";//"basic ";
                authHeader += this.get("token") ? JSON.stringify(this.get("token")) : null;
                return {
                    headers: _.extend(headers, { Authorization: authHeader })
                };
and we do our post calls with restangular this way:

RestangularFactory.all("search/getFields").post(criteria, "");

In our server api we have all post and get calls with:

    [Produces("application/json")]
    [HttpPost]
    public string GetFields([FromBody]CriteriaInputData criteria)
    {
        return ...
    }
If there is any other information that you need please let me know

Thank you very much

harshgMSFT:
@esyorcho can you please post your policy again? Are you sure you are not adding the `Access-Control-Allow-Headers` yourself? This is a response header and not a request header. 
Have you added a `Access-Control-Allow-Headers` in the WithHeaders policy?

Tratcher:
https://github.com/aspnet/CORS/blob/dev/src/Microsoft.AspNet.Cors.Core/CorsService.cs#L147-L154
@harshgMSFT The headers are being added as a list of array values. This results in them being sent as multiple headers. If you need one header they need to be added as a single array element `"value1, value2"`

harshgMSFT:
ah I see .. yeah makes sense...

esyorcho:
Hi @harshgMSFT and @Tratcher ,
I've been trying to add them as 1 header but I couldn't. 
We have in startup application:

app.UseCors(policy => policy.WithOrigins("http://domain1.com", "http://localhost:2025").AllowAnyMethod().AllowAnyHeader().AllowCredentials());

Then we add the authorization header this way:

var authHeader = "";//"basic ";
[...]

PR #23, opened 2015-06-23
Add single csv values so that there is a single response header added…

harshgMSFT:
… per value.

### DIFF
--- a/src/Microsoft.AspNet.Cors.Core/CorsService.cs
+++ b/src/Microsoft.AspNet.Cors.Core/CorsService.cs
@@ -93,8 +93,8 @@ namespace Microsoft.AspNet.Cors.Core
                 return;
             }
 
-            if (!policy.AllowAnyHeader && 
-                requestHeaders != null && 
+            if (!policy.AllowAnyHeader &&
+                requestHeaders != null &&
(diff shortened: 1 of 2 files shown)


harshgMSFT:
@Tratcher

src/Microsoft.AspNet.Cors.Core/CorsService.cs
```diff
@@ -137,7 +137,9 @@ public virtual void ApplyResult(CorsResult result, HttpResponse response)
 
                 if (nonSimpleAllowMethods.Length > 0)
                 {
-                    headers.Add(CorsConstants.AccessControlAllowMethods, nonSimpleAllowMethods);
+                    headers.SetCommaSeparatedValues(
```
Tratcher: If you want them to end up as separate headers you need to call SetValues:
https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNet.Http.Abstractions/IHeaderDictionary.cs#L75

Tratcher: Nevermind, I was remembering the issue backwards.

Tratcher:
#22 :shipit:

PR #27, opened 2015-08-06
Microsoft.AspNet.Cors.Core.Test -> Microsoft.AspNet.Cors.Core.Tests & Microsoft.AspNet.Cors.Test -> Microsoft.AspNet.Cors.Tests

hishamco:
According to the guidelines https://github.com/aspnet/Home/wiki/Engineering-guidelines#solution-and-project-folder-structure-and-naming, we should rename the test project to ```Microsoft.AspNet.Cors.Core.Tests``` instead of ```Microsoft.AspNet.Cors.Core.Test``` and ```Microsoft.AspNet.Cors.Tests``` instead of ```Microsoft.AspNet.Cors.Test```


Eilon:
See https://github.com/aspnet/EventNotification/pull/12#issuecomment-128427434.

PR #29, opened 2015-08-25
Make preflight request header check case-insensitive

Norgerman:
In jquery ajax(maybe in old versions), the `Access-Control-Request-Headers` in `OPTIONS` request value is lower case and will contains `accept` header which is already in `SimpleRequestHeaders`

### DIFF
--- a/src/Microsoft.AspNet.Cors.Core/CorsService.cs
+++ b/src/Microsoft.AspNet.Cors.Core/CorsService.cs
@@ -95,7 +95,8 @@ namespace Microsoft.AspNet.Cors.Core
 
             if (!policy.AllowAnyHeader &&
                 requestHeaders != null &&
-                !requestHeaders.All(header => policy.Headers.Contains(header, StringComparer.Ordinal)))
+                !requestHeaders.All(header => CorsConstants.SimpleRequestHeaders.Contains(header, StringComparer.OrdinalIgnoreCase) ||
+                                              policy.Headers.Contains(header, StringComparer.OrdinalIgnoreCase)))
             {
                 return;
             }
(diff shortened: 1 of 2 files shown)


Eilon:
Logged a bug to track this: https://github.com/aspnet/CORS/issues/32

@kichalla can you review this PR?

src/Microsoft.AspNet.Cors.Core/CorsService.cs
```diff
@@ -97,7 +97,8 @@ public virtual void EvaluatePreflightRequest(HttpContext context, CorsPolicy pol
 
             if (!policy.AllowAnyHeader &&
                 requestHeaders != null &&
-                !requestHeaders.All(header => policy.Headers.Contains(header, StringComparer.Ordinal)))
+                !requestHeaders.All(header => CorsConstants.SimpleRequestHeaders.Contains(header, StringComparer.OrdinalIgnoreCase) ||
+                                              policy.Headers.Contains(header, StringComparer.OrdinalIgnoreCase)))
```
kichalla: Is this change correct? This is supposed to be case-sensitive for any header other than non-simple + content-type header, right?...like can you add a test where the header in the request is "version" and the policy is "Version"...this should not match..

kichalla:
:watch:

src/Microsoft.AspNet.Cors.Core/CorsService.cs
Norgerman: I have tested this use jquery ajax in chrome, `Access-Control-Request-Headers` are always in lower case.

Norgerman:
I have tested this use jquery ajax in chrome, Access-Control-Request-Headers are always in lower case. So the server is not able to get Access-Control-Request-Headers values like Version or verSion, the value will always be version. Thanks for your review.

kichalla:
:shipit:

kichalla:
commit 38728a6bcd11bfe2c5f9c1b17fde1892bf499553

Thanks for the contribution! @Norgerman

Norgerman:
:thumbsup:

Issue #41, opened 2015-10-02
API review updates for CORS


Eilon:
In particular, a lot of types need to move out of the `Cors.Core` namespace. The main `Cors` namespace should have the most commonly used types, and plumbing can go into a sub-namespace.

Eilon:
Met with @kichalla @rynowak and @danroth27 with these conclusions:

- [ ] Merge both packages into `Microsoft.AspNet.Cors`. 
 - The current package separation in practice provided very little value, and unfortunately was self-defeating with respect to the original purpose of having two packages: to encourage developers using MVC to use just the parts MVC needed. Unfortunately it was most common to reference the both packages transitively anyway (which was harmless, but unnecessary)
- [ ] Move `DisableCorsAttribute` and `EnableCorsAttribute` into `.Cors` namespace
- [ ] Move all other types from both original assemblies (including `CorsMiddleware` and `CorsPolicyBuilder`) into `.Infrastructure` namespace.
- [ ] Post announcement that links back to this issue

davidfowl:
Nice

PR #42, opened 2015-10-13
Api review changes

kichalla:
@Eilon 

I will post an announcement once this and reaction changes go in.


src/Microsoft.AspNet.Cors/CorsMiddleware.cs
```diff
@@ -4,7 +4,7 @@
 using System;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Builder;
-using Microsoft.AspNet.Cors.Core;
+using Microsoft.AspNet.Cors.Infrastructure;
 using Microsoft.AspNet.Http;
 using Microsoft.Extensions.Primitives;
```
Eilon: I believe this type is supposed to move to `.Infrastructure`

kichalla: good catch...you are right...also the CorsPolicyBuilder I think...

Eilon: Yeah everything except the two attributes I think. It's in https://github.com/aspnet/CORS/issues/41.

Eilon:
Other than that, :shipit:

kichalla:
475771250bf4f6b8b03ee1c7d30c9f4e4b4816e8

kichalla:
https://github.com/aspnet/CORS/issues/41

PR #48, opened 2015-11-03
Strong name Microsoft.AspNet.Cors.

cesarblum:
https://github.com/aspnet/dnx/issues/2615

cc @davidfowl @Eilon @muratg @moozzyk @pranavkm @Tratcher


Tratcher:
:shipit:

PR #57, opened 2015-12-13
Move functional tests of CORS middleware to this repo

dougbu:
- related to aspnet/Mvc#3612


test/Microsoft.AspNet.Cors.Test/CorsMiddlewareFunctionalTest.cs
```diff
@@ -0,0 +1,98 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.AspNet.Cors.Infrastructure
```
rynowak: Is this the right namespace?

test/Microsoft.AspNet.Cors.Test/CorsTestFixture.cs
```diff
@@ -0,0 +1,145 @@
...
+                    return services.BuildServiceProvider();
+                };
+            }
+
+            _server = TestServer.Create(
+                configureApplication,
+                configureServices: InitializeServices(startupTypeInfo.Assembly, buildServices));
+
+            Client = _server.CreateClient();
+            Client.BaseAddress = new Uri("http://localhost");
```
rynowak: Do we really need all of this? Isn't there something simpler we can do with `TestHost`? /cc @Tratcher

test/WebSites/CorsMiddlewareWebSite/EchoMiddleware.cs
```diff
@@ -0,0 +1,40 @@
...
+        public Task Invoke(HttpContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            context.Response.ContentType = "text/plain; charset=utf-8";
+            var path = context.Request.PathBase + context.Request.Path + context.Request.QueryString;
+            context.Response.WriteAsync(path, Encoding.UTF8);
```
rynowak: `await`

rynowak:
:watch: - mostly to see if we can clean up the startup code.

test/Microsoft.AspNet.Cors.Test/CorsMiddlewareFunctionalTest.cs
dougbu: It matches most of the src classes as well as the other test classes.

test/Microsoft.AspNet.Cors.Test/CorsTestFixture.cs
Tratcher: Yikes. RE: https://github.com/aspnet/Hosting/issues/333. That said, you seem to always use `CorsTestFixture <TStartup>`, so you can remove most of this code and just use `TestServer.UseStartup<TStartup>()`.

dougbu: Most other things we can do w/ `TestHost` run slower because they create numerous instances.

test/WebSites/CorsMiddlewareWebSite/EchoMiddleware.cs
dougbu: Thanx!

test/Microsoft.AspNet.Cors.Test/CorsTestFixture.cs
rynowak: oh noes teh slowz :astonished:  Is this going to cause issues in the cors repo?

dougbu: @Tratcher do you have code you can point me to that uses the `TestServer.UseStartup<TStartup>()` incantation and gets the `ApplicationBasePath` right for the site (not the functional test project)?

rynowak: I guess what I mean is that eventually we hope to align what MVC is doing for these tests with what `TestHost` can do (whatever that means). 

We have two not-super-compatible ways of doing things. 

Can we do these tests the 'standard' way of using `TestHost` and if there's an issue, work to address that separately.

rynowak: > gets the ApplicationBasePath right for the site (not the functional test project)?

What in these tests depends on the base path?

dougbu: What is "standard"? I copied this straight from the MVC repo, changing it only to remove some MVC-specific service additions.

dougbu: > What in these tests depends on the base path?

Maybe nothing but the tests aren't correctly mimicking the production environment if the base path is wrong.

Tratcher: If it's not broke don't fix it.

rynowak: > What is "standard"? I copied this straight from the MVC repo, changing it only to remove some MVC-specific service additions.

`TestHost` is a package we want our users to be able to use. We should be working to address the issues that block us from using it with MVC.

I think it would be great if we didn't copy-paste the code from MVC, especially where the extra things that it does aren't required.

dougbu: > if we didn't copy-paste the code from MVC

This web site and functional test is literally being copy-pasted from MVC. Why are we redesigning how the site is started as part of that move?

Again, I'm not familiar with the `TestServer.UseStartup<TStartup>()`. A couple of tests in this repo use some ugly `TestServer.Create()` calls but that's it. What is this pattern you want me to follow in this fixture?

Tratcher: Oh sorry, `TestServer.Create<TStartup>()`. There's a PR that's refactoring TestServer as we speak but it won't materially change this discussion.

dougbu:
:up::date:

(Startup code greatly cleaned up. Works here though it doesn't appear like it'll work for MVC.)

test/Microsoft.AspNet.Cors.Test/CorsTestFixtureOfT.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    public class CorsTestFixture<TStartup> : IDisposable
+        where TStartup : class
+    {
+        private readonly TestServer _server;
+
+        public CorsTestFixture()
+        {
+            var builder = new WebHostBuilder().UseStartup<TStartup>();
+            _server = new TestServer(builder);
```
dougbu: Couldn't find a `TestServer.Create<TServer>` method so I did this two-step instead.

test/WebSites/CorsMiddlewareWebSite/EchoMiddleware.cs
```diff
@@ -0,0 +1,38 @@
...
+        public Task Invoke(HttpContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            context.Response.ContentType = "text/plain; charset=utf-8";
+            var path = context.Request.PathBase + context.Request.Path + context.Request.QueryString;
+            return context.Response.WriteAsync(path, Encoding.UTF8);
```
dougbu: Method doesn't have to be `async` since this is the last statement.

test/Microsoft.AspNet.Cors.Test/CorsTestFixtureOfT.cs
davidfowl: Much better anyways. We're changing all of the hosting APIs :smile: Stay tuned!

rynowak:
:shipit:

dougbu:
c3c0b45

PR #66, opened 2016-02-28
Fix package metadata

Eilon:
@rynowak @ajaybhargavb


Eilon:
Merged, but please review.

PR #67, opened 2016-02-29
Remove IOptions from CORS

HaoK:
https://github.com/aspnet/Options/issues/122

Are we happy with this Options ctor pattern?  If so I will apply these across the board.  And do something similar for the middlewares which depend on Options as well (add overload)

```
    public CorsOptions() : this(null) { }
    public CorsOptions(IEnumerable<IConfigureOptions<CorsOptions>> configureOptions)
```

cc @divega @davidfowl @lodejard


lodejard:
+1 to make the change across the board

Should the default ctor be used for the defaults, and the enumerable ctor assume the enumerable must not be null? In other words - change which ctor calls the other one?

```cs
public class CorsOptions
{
  public CorsOptions()
  {
    this.MyProperty = "the default";
  }
  public CorsOptions(IEnumerable<IConfigureOptions<CorsOptions>> configs) : this()
  {
    foreach(var config in configs) config.Configure(this);
  }
}
```

What would the middleware ctor overload look like then... Similar to this?
```cs
public class CorsMiddleware
{
  public CorsMiddleware(RequestDelegate next) : this(next, new CorsOptions())
  {
  }
  public CorsMiddleware(RequestDelegate next, CorsOptions options)
  {
    _next = next;
    _options = options;
  }
}
```

HaoK:
Yep I like your suggestion.  Options the default ctor is the main one (with the IEnumerable overload calling this()).  While its the longer with Options ctor that's the main ctor for middleware.

I'll update a repo with a middleware so we can see what that looks like sa well.

HaoK:
See for an example of a middleware only repo

https://github.com/aspnet/Diagnostics/pull/254

HaoK:
If we are more or less happy here, I'll update Security and MVC as the final test to make sure we didn't miss anything before pushing this out

lodejard:
:goberserk: as far as I'm concerned. who else do you think needs to weight in on green-lighting the entire change?

HaoK:
@davidfowl @divega are you guys happy?  I'll fire off a PR in the MVC repo as well to make sure they get a chance to see this trainwreck coming...

lodejard:
:boom:  :steam_locomotive: :train: :train: :train: :see_no_evil:

divega:
I am fine with this.

davidfowl:
Very nice. Can we try this pattern with a few 3rd party DI systems? Would be good to add this to our spec tests

HaoK:
@davidfowl Isn't this ctor pattern already covered by this test?

https://github.com/aspnet/DependencyInjection/blob/dev/src/Microsoft.Extensions.DependencyInjection.Specification.Tests/DependencyInjectionSpecificationTests.cs#L590

halter73:
Is there any DI systems other than our own and Autfac that actually run against those specification tests?

PR #76, opened 2016-04-27
Ability to allow cors on wildcard subdomain matches like so...

sjdirect:
-Ability to allow cors on wildcard subdomain matches like so...
-policy.Origins.Add("http://*.example.com"); //Allows all sudomains of example.com

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -1,7 +1,8 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
-    <add key="AspNetVNext" value="https://www.myget.org/F/aspnetcirelease/api/v3/index.json" />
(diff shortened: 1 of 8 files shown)


Eilon:
Closing because we ended up merging https://github.com/aspnet/CORS/pull/94. Thanks!

PagesByZ:
Where do I add that?

[EnableCors(origins: "http://localhost:64, http://mysite:100, http://appset", headers: "*", methods: "*")] works for me but when I do the following, [EnableCors(origins: "http://localhost:64, http://mysite:100, http://appset, http://appset/mysubdomain", headers: "*", methods: "*")] nothing works.

How can I fix it please. Thanks.

Eilon:
@kichalla can you help out?

MienDev:
@PagesByZ  `http://appset/mysubdomain` is not a valid origin.   [rfc6454](https://tools.ietf.org/html/rfc6454) for your reference.

PR #84, opened 2016-08-17
Include VaryByOrigin when there are multiple Origins configured

michielpost:
The Vary: Origin header is currently only returned when AllowAnyOrigin and SupportsCredentials is true
When AllowAnyOrigin is not set and there is a origin match for the request, the AllowedOrigin is returned but without the VaryByOrigin header.
That's not a problem when only 1 Origin is configured, but when there are multiple Origins configured, the VaryByOrigin should also be returned.

### DIFF
--- a/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
+++ b/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
@@ -245,6 +245,11 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
             else if (policy.Origins.Contains(origin))
             {
                 result.AllowedOrigin = origin;
+
+                if(policy.Origins.Count > 1)
+                {
+                    result.VaryByOrigin = true;
+                }
             }
         }
 
--- a/test/Microsoft.AspNetCore.Cors.Test/CorsServiceTests.cs
+++ b/test/Microsoft.AspNetCore.Cors.Test/CorsServiceTests.cs
@@ -881,6 +881,24 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
             Assert.Equal("30", httpContext.Response.Headers["Access-Control-Max-Age"]);
         }
 
+        [Fact]
+        public void EvaluatePolicy_MultiOriginsPolicy_ReturnsVaryByOriginHeader()
+        {
+            // Arrange
+            var corsService = new CorsService(new TestCorsOptions());
+            var requestContext = GetHttpContext(origin: "http://example.com");
+            var policy = new CorsPolicy();
+            policy.Origins.Add("http://example.com");
+            policy.Origins.Add("bar");
+
+            // Act
+            var result = corsService.EvaluatePolicy(requestContext, policy);
+
+            // Assert
+            Assert.NotNull(result.AllowedOrigin);
+            Assert.True(result.VaryByOrigin);
+        }
+
 
 
         private static HttpContext GetHttpContext(
--- a/test/Microsoft.AspNetCore.Cors.Test/CorsServiceTests.cs
+++ b/test/Microsoft.AspNetCore.Cors.Test/CorsServiceTests.cs
@@ -899,26 +899,26 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
             Assert.True(result.VaryByOrigin);
         }
 
-		[Fact]
-		public void EvaluatePolicy_MultiOriginsPolicy_NoMatchingOrigin_ReturnsInvalidResult()
-		{
-			// Arrange
-			var corsService = new CorsService(new TestCorsOptions());
-			var requestContext = GetHttpContext(origin: "http://example.com");
-			var policy = new CorsPolicy();
-			policy.Origins.Add("http://example-two.com");
-			policy.Origins.Add("http://example-three.com");
+        [Fact]
+        public void EvaluatePolicy_MultiOriginsPolicy_NoMatchingOrigin_ReturnsInvalidResult()
+        {
+            // Arrange
+            var corsService = new CorsService(new TestCorsOptions());
+            var requestContext = GetHttpContext(origin: "http://example.com");
+            var policy = new CorsPolicy();
+            policy.Origins.Add("http://example-two.com");
+            policy.Origins.Add("http://example-three.com");
 
-			// Act
-			var result = corsService.EvaluatePolicy(requestContext, policy);
+            // Act
+            var result = corsService.EvaluatePolicy(requestContext, policy);
 
-			// Assert
-			Assert.Null(result.AllowedOrigin);
-			Assert.False(result.VaryByOrigin);
-		}
+            // Assert
+            Assert.Null(result.AllowedOrigin);
+            Assert.False(result.VaryByOrigin);
+        }
 
 
-		private static HttpContext GetHttpContext(
+        private static HttpContext GetHttpContext(
             string method = null,
             string origin = null,
             string accessControlRequestMethod = null,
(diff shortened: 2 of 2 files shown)


michielpost:
Any feedback on this PR?

Eilon:
@kichalla @jbagga thoughts on this?

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
```diff
@@ -245,6 +245,11 @@ private void AddOriginToResult(string origin, CorsPolicy policy, CorsResult resu
             else if (policy.Origins.Contains(origin))
             {
                 result.AllowedOrigin = origin;
+
+                if(policy.Origins.Count > 1)
```
kichalla: It appears even with single origin, this should be set
https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS
 
image

jbagga:
The [CORS spec](https://www.w3.org/TR/cors/) mentions that for _multiple_ origins the resource should send the Vary header

> Resources that wish to enable themselves to be shared with multiple Origins... should send a Vary: Origin HTTP header or provide other appropriate control directives to prevent caching of such responses, which may be inaccurate if re-used across-origins.

I think the PR is a reasonable addition. @kichalla can elaborate on current behavior.

Eilon:
@jbagga @kichalla thanks! Can one of you log a bug detailing what doesn't work, so we can associate with this PR?

test/Microsoft.AspNetCore.Cors.Test/CorsServiceTests.cs
```diff
@@ -881,6 +881,24 @@ public void ApplyResult_PreflightMaxAge_MaxAgeHeaderAdded()
             Assert.Equal("30", httpContext.Response.Headers["Access-Control-Max-Age"]);
         }
 
+        [Fact]
+        public void EvaluatePolicy_MultiOriginsPolicy_ReturnsVaryByOriginHeader()
```
kichalla: Update this test to have variations: one origin, multiple origins.
Add a negative test case where multiple origins and request from non-allowable origin doesn't have origin header

jbagga:
Addresses #97

Eilon:
Thanks! Assigned to @jbagga (incl. the bug) to do final review and merging.

jbagga:
@michielpost Thank you for the PR. There are some changes @kichalla suggested. We can then review and merge it in!

test/Microsoft.AspNetCore.Cors.Test/CorsServiceTests.cs
```diff
@@ -881,6 +881,24 @@ public void ApplyResult_PreflightMaxAge_MaxAgeHeaderAdded()
...
+        {
+            // Arrange
+            var corsService = new CorsService(new TestCorsOptions());
+            var requestContext = GetHttpContext(origin: "http://example.com");
+            var policy = new CorsPolicy();
+            policy.Origins.Add("http://example.com");
+            policy.Origins.Add("bar");
```
michielpost: Ok, changed the origin to a valid url

test/Microsoft.AspNetCore.Cors.Test/CorsServiceTests.cs
michielpost: I believe One origin is handled in other test cases, like EvaluatePolicy_NoMatchingOrigin_ReturnsInvalidResult
This test case handles Multiple Origins.

And I've added the negative test case EvaluatePolicy_MultiOriginsPolicy_NoMatchingOrigin_ReturnsInvalidResult

test/Microsoft.AspNetCore.Cors.Test/CorsServiceTests.cs
```diff
@@ -881,9 +881,44 @@ public void ApplyResult_PreflightMaxAge_MaxAgeHeaderAdded()
...
+            // Assert
+            Assert.NotNull(result.AllowedOrigin);
+            Assert.True(result.VaryByOrigin);
+        }
+
+		[Fact]
+		public void EvaluatePolicy_MultiOriginsPolicy_NoMatchingOrigin_ReturnsInvalidResult()
+		{
+			// Arrange
+			var corsService = new CorsService(new TestCorsOptions());
```
Eilon: Looks like the latest commit inserted tabs instead of spaces - can you change it back to all spaces?

Eilon: So much nicer, thanks! 😄

jbagga:
Looks good! Once @kichalla has approved, I will merge this in.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
andrew-laughlin: I believe this is correct. Even if a single origin is returned, the `Vary: origin` header should be set on the response to allow HTTP caches to respond correctly. Chrome in particular will return cross-origin requests from cache if this header is not set. This causes errors, for example, if applications in separate tabs make CORS requests to the same resource. Chrome will cache the first request -- which contains the header `Access-Control-Allow-Origin: http://origin1` but not `Vary: origin` -- and serve the cached response up to the second tab, which will fail because it's origin doesn't match the `Access-Control-Allow-Origin` header.

For discussion see: https://bugs.chromium.org/p/chromium/issues/detail?id=409090

#https://github.com/IdentityModel/oidc-client-js/issues/275

jbagga: @kichalla @Eilon we may need to discuss this and possibly change this to `policy.Origins.Count >= 1`. The CORS spec does not mention the header being set for a single origin but it looks like that's what Chrome expects for caching to correctly work as @andrew-laughlin explained.

Eilon: @jbagga - I guess log a new bug on that?

jbagga: Filed here: #111

jbagga: @andrew-laughlin Do you have a repro sample? When making the same request from the second tab, are you using a different origin? 

> serve the cached response up to the second tab, which will fail because it's origin doesn't match the Access-Control-Allow-Origin header.

Do you mean the Response is not empty as it should be for an origin not in Access-Control-Allow-Origin? Failing because the origin is not in Access-Control-Allow-Origin is expected. I am not sure what you mean by "serve the cached response...which will fail".
cc @kichalla

jbagga: Interestingly, the [MDN doc](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS) has been modified to say that 

> If the server specifies an origin host rather than "*", **then it could also include Origin in the Vary response header** to indicate to clients that server responses will differ based on the value of the Origin request header.

So it isn't required.

nickwender: @jbagga, was that page really updated? I looked through the history and could not find such a modification.

Interestingly,  the reference page for the `Access-Control-Allow-Origin` header still says that `vary: origin` must be set: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin#CORS_and_caching

Perhaps, re-opening https://github.com/aspnet/CORS/issues/111 should be considered?

image

Eilon: @nickwender - can you file a new issue for this? It's a bit hard to follow the discussion in this old PR.

PR #85, opened 2016-08-30
WildCard support in origins

alireza-naghizadeh:
Ability to allow using wildcard in origins...
Example (like DNS & IIS.10 bindings):

```
app.UseCors(policy => policy
    policy.WithOrigins("http://example.com", "http://*.example.com")
);
```

@dnfclas

### DIFF
--- a/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicy.cs
+++ b/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicy.cs
@@ -3,9 +3,7 @@
 
 using System;
 using System.Collections.Generic;
-using System.Collections.ObjectModel;
 using System.Text;
-using System.Text.RegularExpressions;
 
 namespace Microsoft.AspNetCore.Cors.Infrastructure
 {
@@ -82,13 +80,7 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
         /// <summary>
         /// Gets the origins that are allowed to access the resource.
         /// </summary>
-        public ObservableCollection<string> Origins { get; } = new ObservableCollection<string>();
-
-        /// <summary>
-        /// Gets the regular origins that are allowed to access the resource.
-        /// </summary>
-        public IList<Regex> RegularOrigins { get; } = new List<Regex>();
-
+        public IList<string> Origins { get; } = new List<string>();
 
         /// <summary>
         /// Gets or sets the <see cref="TimeSpan"/> for which the results of a preflight request can be cached.
@@ -148,24 +140,6 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
             builder.Append(string.Join(",", ExposedHeaders));
             builder.Append("}");
             return builder.ToString();
-        }
-
-        /// <summary>
-        /// Creates a new instance of the <see cref="CorsPolicy"/>.
-        /// </summary>
-        public CorsPolicy()
-        {
-            // Regenerate 'RegularOrigins' every time the 'Origins' has changed.
-            Origins.CollectionChanged += (sender, e) =>
-            {
-                RegularOrigins.Clear();
-
-                foreach (var origin in Origins)
-                {
-                    var pattern = $"^{Regex.Escape(origin).Replace(@"\*", ".*")}$";
-                    RegularOrigins.Add(new Regex(pattern));
-                }
-            };
         }
     }
 }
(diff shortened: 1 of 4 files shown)


alireza-naghizadeh:
The better example (like DNS & IIS.10 bindings):

app.UseCors(policy => policy
policy.WithOrigins("http://example.com", "http://*.example.com")
);

@dnfclas

danroth27:
Confirmed the CLA was signed

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
```diff
@@ -82,7 +82,7 @@ public CorsResult EvaluatePolicy(HttpContext context, CorsPolicy policy)
         public virtual void EvaluateRequest(HttpContext context, CorsPolicy policy, CorsResult result)
         {
             var origin = context.Request.Headers[CorsConstants.Origin];
-            if (StringValues.IsNullOrEmpty(origin) || !policy.AllowAnyOrigin && !policy.Origins.Contains(origin))
+            if (StringValues.IsNullOrEmpty(origin) || !policy.AllowAnyOrigin && !policy.RegularOrigins.Any(r => r.IsMatch(origin)))
```
troydai: What's the performance implication here? The impact could be big.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicy.cs
```diff
@@ -141,5 +149,23 @@ public override string ToString()
...
             return builder.ToString();
         }
+
+        /// <summary>
+        /// Creates a new instance of the <see cref="CorsPolicy"/>.
+        /// </summary>
+        public CorsPolicy()
+        {
+            // Regenerate 'RegularOrigins' every time the 'Origins' has changed.
+            Origins.CollectionChanged += (sender, e) =>
```
troydai: Are policies all generated and set during configuration? If so using a `ObservableCollection` seems to be an overkill. Are there mechanism to update patterns once when all the origins are added?

alireza-naghizadeh: > Are policies all generated and set during configuration?

No! We can change origins out of the configuration... When you need dynamic origins, you can write a custom service divided from `CorsService`... I have a project that I change origins dynamically in `EvaluateRequest` & `EvaluatePreflightRequest` with access to `policy.Origins`...
So, by using `ObservableCollection` we can generate `RegularOrigins` each time is needed.

davidfowl: We shouldn't use an observable collection, instead an interface with a method call would be better than having code react to an event handlers whenever the collection changes.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
davidfowl: Regex by default? I don't think so 😄

alireza-naghizadeh: We can improve performance by check both (combined), by default check origin strings and if needed wildcard then check regular origins...

davidfowl: Sure, anything with regex raises a red flag. We avoid them like the plague and performance results need to be shown whenever they are used. They also need a timeout so @blowdart doesn't freak out 😄

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicy.cs
davidfowl: I also don't see how this is related to the PR. Can you make this change only about wildcards and separate dynamic configuration in another issue/PR.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
davidfowl: Actually why do you even need to support regex? Just do the wildcard matching by hand, it'll be more predictable and efficient that way. Also we can start by allowing less things. Opening it up to regex makes it hard to optimize.

alireza-naghizadeh:
I changed the codes! Now there isn't any Regex...
I was solving many issues (dynamic origins, regular support...), but @davidfowl is right, Now we should solve this issue (WildCard support)...
Thanks

Eilon:
We merged https://github.com/aspnet/CORS/pull/94, so closing this.

PR #93, opened 2016-10-30
Origins starting with // is translated to two origins (http and https)

moander:


src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyBuilder.cs
```diff
@@ -40,7 +40,15 @@ public CorsPolicyBuilder WithOrigins(params string[] origins)
         {
             foreach (var req in origins)
             {
-                _policy.Origins.Add(req);
+                if (req.StartsWith("//"))
```
Eilon: Is there a reason you'd use this format here?

Eilon:
Closing this because at least for the moment there doesn't seem to be much demand for this feature. To revive the discussion please start by logging a new issue and we can gauge the interest level of this feature.

Issue #89, opened 2016-10-05
Allow a more flexible AllowOrigin response


mharthoorn:
Instead of the CorsService peeking into the CorsPolicy

``` csharp
   policy.Origins.Contains(origin)
```

A method `OriginAllowed(string origin)` would allow for a much more flexible lookup policy. The service would then call:

``` csharp
  policy.OriginAllowed(origin)
```

This allows the policy to respond with a more ad hoc response like a (database) lookup.
The default would simply be:

``` csharp
        public virtual bool OriginAllowed(string origin)
        {
            return this.Origins.Contains(origin);
        }
```

But an override policy could become something like: 

``` csharp
        public override bool OriginAllowed(string origin)
        {
            return repository.Exists<Sites>(s => s.Origin == origin);
        }
```

Choc13:
See issue #75

Eilon:
We will consider this for the next version.

aspnet-hello:
This issue was moved to aspnet/Home#2379

Issue #75, opened 2016-04-26
Cannot find documentation/examples of how to support wildcard subdomains


sjdirect:
Below is what i have posted on  [stackoverflow](http://stackoverflow.com/questions/36877652/configure-cors-to-allow-all-subdomains-using-asp-net-core-asp-net-5-mvc6-vnex) regarding this issue... 

I have cors setup correctly in an ASP.NET Core web app. Im using the following package...

```
"Microsoft.AspNet.Cors": "6.0.0-rc1-final"
```

and here is the startup.cs snippet...

```
public virtual IServiceProvider ConfigureServices(IServiceCollection services)
{
    services.AddCors
    (
        options =>
        {
            options.AddPolicy
            (
                CORSDefaults.PolicyName, 
                builder =>
                {
                    //From config...
                    var allowedDomains = new []{"http://aaa.somewhere.com","https://aaa.somewhere.com","http://bbb.somewhere.com","https://bbb.somewhere.com"};

                    //Load it
                    builder
                        .WithOrigins(allowedDomains)
                        .AllowAnyHeader()
                        .AllowAnyMethod()
                        .AllowCredentials();
                }
            );
        }
    );
}
```

This works great except that the list of subdomains to allow is growing fast and I want to allow all subdomains of "somewhere.com". Something like "*.somewhere.com". I cant seem to find any documentation on how to do this in the new ASP.NET Core (MVC6, ASP.NET5, VNext). All the docs/examples I'm finding that demonstrate how to do this are for earlier versions of MVC or WebApi. How can I achieve this in the new stack?

sjdirect:
Submitted this pull request on the master branch...

https://github.com/aspnet/CORS/pull/76

blowdart:
I don't believe it is supported.

rynowak:
@blowdart - I think the question is **should** we support it... If I'm missing something can you provide more info?

blowdart:
I saw the PR. It feels wrong from a security perspective, the CORS spec doesn't mention wildcards at all, you ought to list all your domains, especially when you know someone will whitelist azurewebsites.net because they don't understand the risks.

However it does make everything easier, so from a user PoV it's a good thing.

Do rails et al support wildcards?

sjdirect:
@blowdart, what if I wanted to provide an api only to azure users? Without this functionality I'm not sure how it could be done without some heavy initialization costs and data availability from Azure.

blowdart:
[...]

PR #94, opened 2016-11-01
Add support for configuring CorsPolicy with method to determine whether an origin is allowed

Choc13:
As per discussions in #75 and #89 this PR add an `IsOriginAllowed` property to the `CorsPolicy`. This allows clients the ability to configure the policy with an arbitrary function that evaluates whether a given origin should be allowed.

Also added `CorsPolicyBuilder.AllowWildcardSubdomains` which configures the `CorsPolicy` with a `IsOriginAllowed` function that checks whether the given origin is a subdomain of any origins on the policy that contain a wildcarded subdomain.


src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyBuilder.cs
```diff
@@ -91,6 +91,18 @@ public CorsPolicyBuilder WithMethods(params string[] methods)
         }
 
         /// <summary>
+        /// Adds the specified <paramref name="isOriginAllowed"/> to the policy.
+        /// </summary>
+        /// <param name="isOriginAllowed">The function used by the policy to evaluate if an origin is allowed.</param>
+        /// <returns>The current policy builder</returns>
+        public CorsPolicyBuilder WithIsOriginAllowed(Func<string, bool> isOriginAllowed)
+        {
+            _policy.IsOriginAllowed = isOriginAllowed;
```
Eilon: The `WithXyz` methods all *combine* logic, whereas this method is a *destructive* method that replaces a setting. Would a better name for this method be something like `SetIsOriginAllowed(...)`?

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyBuilder.cs
```diff
@@ -144,6 +156,17 @@ public CorsPolicyBuilder AllowAnyHeader()
         }
 
         /// <summary>
+        /// Enables the policy to be configured with wildcarded domains that will then allow
+        /// CORS from any matching subomdain
+        /// </summary>
+        /// <returns>The current policy builder</returns>
+        public CorsPolicyBuilder AllowWildcardSubdomains()
+        {
+            _policy.IsOriginAllowed = _policy.IsOriginAnAllowedSubdomain;
```
Eilon: Related to the above comment, this method is also *destructive*. Would it make more sense to have some way of *combining* the `Func<>`s? That would certainly be more flexible, though perhaps too complex? Thoughts?

src/Microsoft.AspNetCore.Cors/Infrastructure/UriExtensions.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Cors.Infrastructure
+{
+    internal static class UriExtensions
+    {
+        public static bool IsSubdomainOf(this Uri subdomain, Uri domain)
```
Eilon: We generally avoid extension methods in cases like this - we prefer just a plain old static helper method.

src/Microsoft.AspNetCore.Cors/Infrastructure/UriExtensions.cs
```diff
@@ -0,0 +1,18 @@
...
+        {
+            return subdomain.IsAbsoluteUri 
+                && domain.IsAbsoluteUri
+                && subdomain.Scheme == domain.Scheme
+                && subdomain.Host.EndsWith(domain.Host);
```
Eilon: 1. I don't think this check is right: `"example.com".EndsWith("ample.com") == true`. Does this logic need to be a bit smarter about being a "subdomain"?
2. Does this also need to check things like ports? Is that a factor in CORS calculation?

test/Microsoft.AspNetCore.Cors.Test/UriExtensionsTests.cs
```diff
@@ -0,0 +1,36 @@
...
+            Assert.Equal(subdomain.IsSubdomainOf(domain), expectedResult);
+        }
+
+        public static IEnumerable<object[]> IsSubdomainOfTestData
+        {
+            get
+            {
+                return new[]
+                {
+                    new object[] {new Uri("http://sub.domain"), new Uri("http://domain"), true},
```
Eilon: Per my previous comment, I'm guessing that this:

```c#
new object[] {new Uri("http://subdomain"), new Uri("http://domain"), true}, 
```

Would also pass, even though it is wrong. (`subdomain` is clearly not a sub-domain of `domain`.)

src/Microsoft.AspNetCore.Cors/Infrastructure/UriExtensions.cs
Choc13: Happy to change this to a static helper.

src/Microsoft.AspNetCore.Cors/Infrastructure/UriExtensions.cs
Choc13: Yes. Good spot. I'll add the additional test case and tighten up the constraints to partion the subdomains and look for an exact match

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyBuilder.cs
Choc13: I think combing the funcs could make this too complex. For instance should funcs be logically combined with AND or OR? 

I was in 2 minds about whether or not to provide this implementation, but figured it was better to have a robust implementation in here that was well tested to avoid others from making mistakes that might make their app insecure. However if people want to use this to different conditions they should probably take complete control of that logic to make it easier to understand. 

I will try and think of a better name for this method.

Eilon:
Great, thanks, looking forward to see your updates!

Choc13:
Hi @Eilon,

I've made all the changes that you requested. 

Thanks,

Matt

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyBuilder.cs
```diff
@@ -148,14 +148,37 @@ public CorsPolicyBuilder AllowAnyHeader()
...
+            return this;
+        }
+
+        /// <summary>
+        /// Sets the <see cref="CorsPolicy.IsOriginAllowed"/> property of the policy to be configured with 
+        /// a function that allows origins to match configured wildcarded domain when evaluating if the 
+        /// origin is allowed.
+        /// </summary>
+        /// <returns>The current policy builder</returns>
+        public CorsPolicyBuilder SetIsOriginAllowedToAllowWildcardSubdomains()
```
Eilon: Nice name 😄

src/Microsoft.AspNetCore.Cors/Infrastructure/UriHelpers.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    internal static class UriHelpers
+    {
+        public static bool IsSubdomainOf(Uri subdomain, Uri domain)
+        {
+            return subdomain.IsAbsoluteUri 
+                && domain.IsAbsoluteUri
+                && subdomain.Scheme == domain.Scheme
+                && subdomain.Port == domain.Port
+                && subdomain.Host.EndsWith($".{domain.Host}");
```
Eilon: Is the purpose of the subdomain check to allow *only* subdomains, or also to allow the domain itself? This updated check will fail for the domain itself (i.e. a domain is not a subdomain of itself).

I have a feeling that the more common case is to allow both subdomains and the domain itself. Thoughts?

src/Microsoft.AspNetCore.Cors/Properties/AssemblyInfo.cs
```diff
@@ -3,9 +3,16 @@
...
 using System.Reflection;
 using System.Resources;
+using System.Runtime.CompilerServices;
 
 [assembly: AssemblyMetadata("Serviceable", "True")]
 [assembly: NeutralResourcesLanguage("en-US")]
 [assembly: AssemblyCompany("Microsoft Corporation.")]
 [assembly: AssemblyCopyright("© Microsoft Corporation. All rights reserved.")]
 [assembly: AssemblyProduct("Microsoft ASP.NET Core")]
+[assembly: InternalsVisibleTo("Microsoft.AspNetCore.Cors.Test,PublicKey="+
```
Eilon: Can have this all on one line (yeah, it's a super long line, but it's not a very interesting one!).

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicy.cs
```diff
@@ -15,6 +15,14 @@ public class CorsPolicy
         private TimeSpan? _preflightMaxAge;
 
         /// <summary>
+        /// Default constructor for a CorsPolicy
```
Eilon: End all sentences with periods.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicy.cs
```diff
@@ -15,6 +15,14 @@ public class CorsPolicy
...
+        /// </summary>
+        public CorsPolicy()
+        {
+            this.IsOriginAllowed = DefaultIsOriginAllowed;
```
Eilon: Remove `this.`

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicy.cs
Eilon: (Applies to lots of doc comments in this PR.)

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+            var originUri = new Uri(origin, UriKind.Absolute);
+            return policy.Origins
+                .Where(o => o.Contains(_WildcardSubdomain))
+                .Select(CreateDomainUri)
+                .Any(domain => UriHelpers.IsSubdomainOf(originUri, domain));
+        }
+
+        private static Uri CreateDomainUri(string origin)
+        {
+            return new Uri(origin.Replace(_WildcardSubdomain, ""), UriKind.Absolute);
```
Eilon: Use `string.Empty`

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public static bool IsOriginAnAllowedSubdomain(this CorsPolicy policy, string origin)
+        {
+            if (policy.Origins.Contains(origin))
+            {
+                return true;
+            }
+            var originUri = new Uri(origin, UriKind.Absolute);
+            return policy.Origins
+                .Where(o => o.Contains(_WildcardSubdomain))
```
Eilon: Should this be `.Contains` or `.StartsWith`? (Should also pass in `StringComparison.Ordinal`.)

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyBuilder.cs
```diff
@@ -148,14 +148,37 @@ public CorsPolicyBuilder AllowAnyHeader()
...
+        /// <returns>The current policy builder</returns>
+        public CorsPolicyBuilder SetIsOriginAllowed(Func<string, bool> isOriginAllowed)
+        {
+            _policy.IsOriginAllowed = isOriginAllowed;
+            return this;
+        }
+
+        /// <summary>
+        /// Sets the <see cref="CorsPolicy.IsOriginAllowed"/> property of the policy to be configured with 
+        /// a function that allows origins to match configured wildcarded domain when evaluating if the
```
Eilon: Change to: `... allows origins to match a configured wildcard domain when evaluating ...`

Eilon:
@jbagga @kichalla, I've done some initial reviewing of this PR, but you two are the experts, so please take a look as well.

cc @blowdart because this is CORS, and CORS is security, and security is you 😄

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyExtensions.cs
Choc13: I believe the `policy.Origins` are configured with domains that include the protocol, such as `http://example.com` therefore I opted for `Contains`. However if you are concerned that this might allow someone to configure the policy with `http://sub.*.domain.com`, which would get turned into a Uri like `http://sub.domain.com` then I could alter the logic here. I guess it could be changed to instead look for `://*.` so that the wildcard part is just after the protocol.

Yes will add the `StringComparison.Ordinal` argument.

src/Microsoft.AspNetCore.Cors/Infrastructure/UriHelpers.cs
blowdart: This is pretty much how X509 certificates and HTTPS matching works. If you get a wildcard cert for *.contoso.com it doesn't match contoso.com. Mirroring this behaviour seems fine to me.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyExtensions.cs
blowdart: Again, if we match HTTPS, only wildcards at the beginning should apply. Which seems reasonable to me.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+
+namespace Microsoft.AspNetCore.Cors.Infrastructure
+{
+    internal static class CorsPolicyExtensions
+    {
+        private static string _WildcardSubdomain = "*.";
```
blowdart: const

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicy.cs
```diff
@@ -141,5 +154,10 @@ public override string ToString()
             builder.Append("}");
             return builder.ToString();
         }
+
+        private bool DefaultIsOriginAllowed(string origin)
+        {
+            return Origins.Contains(origin);
```
blowdart: Needs StringComparison.OrdinalIgnoreCase? or at least Ordinal.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyExtensions.cs
Choc13: @Eilon, according to the [docs](https://msdn.microsoft.com/en-us/library/dy85x1sa(v=vs.110).aspx), `string.Contains` does not take an `StringComparison` argument. But it does perform an ordinal case-sensitive comparison anyway when searching for the substring. Are you happy with this?

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicy.cs
Choc13: Done, as far as I can see.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicy.cs
Choc13: Done.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyExtensions.cs
Choc13: Done.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyExtensions.cs
Choc13: Done.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyBuilder.cs
Choc13: Done.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicy.cs
Choc13: @blowdart. Happy to add this, but note I lifted this straight out of the `CorsService`. This was the logic that was previously being used to evaluate where an origin is allowed, but obviously this has now been replaced with a call to the `IsOriginAllowed` func on the policy and I am just setting this as the default behaviour. So this would potentially have a subtle change on the default behaviour for checking origins.

Choc13: Changing to `OridinalIgnoreCase` also breaks the following test:

`Microsoft.AspNetCore.Cors.Infrastructure.CorsServiceTests.TryValidateOrigin_DoesCaseSensitiveComparison`

So for now I am going to set it to `Ordinal`. Unless you are happy for me to change this test case?

src/Microsoft.AspNetCore.Cors/Infrastructure/UriHelpers.cs
Eilon: Got it, this makes perfect sense to me. Thanks!

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyExtensions.cs
Eilon: I believe this is acceptable. Thanks for doing the research!

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicy.cs
```diff
@@ -15,6 +16,14 @@ public class CorsPolicy
         private TimeSpan? _preflightMaxAge;
 
         /// <summary>
+        /// Default constructor for a CorsPolicy.
+        /// </summary>
+        public CorsPolicy()
+        {
+            IsOriginAllowed = DefaultIsOriginAllowed;
```
Eilon: Wondering - does the compiler allow initializing this assignment on the property itself? (C#6 property initializers)

jbagga:
Looks good to me! :shipit:

Eilon:
@jbagga if you and @kichalla are good with this, please merge the PR.

kichalla:
Thanks! for the PR @Choc13. Will merge it today.

kichalla:
8214954d5b0dafe191890da9a3de86f5b607ea69

Issue #78, opened 2016-05-09
Cors needs logging


rynowak:
CORS really badly needs some logging - this is an area of frequent question-asking. We should try to automate our question answering.

Ideas:
- Whether or not request was preflight (debug)
- Where or not request had origin header (debug)
- Policy success or failure with failure reason (info)

PR #95, opened 2016-11-17
Sample and logging for CORS

jbagga:
Addresses #69 and #78 

Need suggestions for log messages (add more, modify current messages, change log location etc.) and if the sample is too simple/needs ajax instead of javascript. The sample will need a README file (will be added before checking in).

cc @rynowak @Tratcher


SampleDestination/Program.cs
```diff
@@ -0,0 +1,31 @@
...
+using Microsoft.Extensions.Configuration;
+
+namespace SampleDestination
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            var config = new ConfigurationBuilder()
+        .SetBasePath(Directory.GetCurrentDirectory())
```
Tratcher: Remove the config file

SampleDestination/project.json
```diff
@@ -0,0 +1,50 @@
+{
+  "dependencies": {
+    "Microsoft.NETCore.App": {
+      "version": "1.1.0-*",
+      "type": "platform"
+    },
+    "Microsoft.AspNetCore.Diagnostics": "1.2.0-*",
```
Tratcher: Remove, this adds a new repo dependency.
http://aspnetci/viewLog.html?buildId=153650&buildTypeId=UniverseCoherence&tab=buildLog&state=248%2C240#_state=248,240&focus=570

SampleDestination/StatusMiddleware.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <param name="context">The <see cref="HttpContext"/> of the current request.</param>
+        /// <returns>A <see cref="Task"/> that completes when writing to the response is done.</returns>
+        public Task Invoke(HttpContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+                 
+            return context.Response.WriteAsync(context.Response.StatusCode.ToString());
```
Tratcher: This can be inlined in Startup.Configure. See app.Run

SampleDestination/hosting.json
```diff
@@ -0,0 +1,3 @@
+{
+  "server.urls": "http://destination.sample.com:80"
```
Tratcher: Move to Program.Main UseUrls.

Tratcher: The host name doesn't matter here, use *.
Also, 80 is likely to cause a conflict. try `http://*:5000/`

SampleOrigin/project.json
```diff
@@ -0,0 +1,50 @@
...
+  "dependencies": {
+    "Microsoft.NETCore.App": {
+      "version": "1.1.0-*",
+      "type": "platform"
+    },
+    "Microsoft.AspNetCore.Diagnostics": "1.2.0-*",
+    "Microsoft.AspNetCore.Server.IISIntegration": "1.2.0-*",
+    "Microsoft.AspNetCore.Server.Kestrel": "1.2.0-*",
+    "Microsoft.Extensions.Logging.Console": "1.2.0-*",
+    "Microsoft.AspNetCore.StaticFiles": "1.2.0-*",
```
Tratcher: Can this be removed? This is another new repo dependency, but it's negotiable.

SampleOrigin/wwwroot/Index.html
```diff
@@ -0,0 +1,49 @@
...
+
+            request.send();
+        }
+
+    </script>
+    <title></title>
+</head>
+<body>
+    <form method="post">
+            <input id="clickMe" type="button" value="CORS Request" onclick="makeCORSRequest('GET');" />
```
kichalla: This is a bigger ask but it would be nice if the details of the CORS request was dynamic in nature. I say this because setting up a repro for CORS (whenever a user says that some scenario is not working) is a pain so if we could add a bit of dynamic nature to this sample it would be very helpful for us, internally, to quickly validate issues filed by users. By dynamic, I mean for example, selecting the http method via a drop down or posting data of different content-types etc. Please check with @Eilon regarding this.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
```diff
@@ -82,24 +97,41 @@ public CorsResult EvaluatePolicy(HttpContext context, CorsPolicy policy)
         public virtual void EvaluateRequest(HttpContext context, CorsPolicy policy, CorsResult result)
         {
             var origin = context.Request.Headers[CorsConstants.Origin];
-            if (StringValues.IsNullOrEmpty(origin) || !policy.AllowAnyOrigin && !policy.Origins.Contains(origin))
+            if (StringValues.IsNullOrEmpty(origin))
             {
                 return;
```
kichalla: we would like to log in this case too?

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
```diff
@@ -82,24 +97,41 @@ public CorsResult EvaluatePolicy(HttpContext context, CorsPolicy policy)
...
             }
 
+            if (!policy.AllowAnyOrigin && !policy.Origins.Contains(origin))
+            {
+                _logger?.RequestHasOriginHeader();
+                _logger.PolicyFailure($"Request origin {origin} does not have permission to access the resource.");
```
kichalla: Cors specification is strict when matching entries. I think its better to add a brief text about it in the messages, otherwise users tend to overlook the tiny details.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
jbagga: Will it still be a policy failure though? Or an exception that the request had no origin?

src/Microsoft.AspNetCore.Cors/Internal/CORSLoggerExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+                "The request has an origin header.");
+
+            _policySuccess = LoggerMessage.Define(
+                LogLevel.Information,
+                3,
+                "Policy execution successful.");
+
+            _policyFailure = LoggerMessage.Define<string>(
+                LogLevel.Information,
+                3,
```
kichalla: Change this event id

SampleDestination/Startup.cs
```diff
@@ -0,0 +1,36 @@
...
+        public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory)
+        {
+            loggerFactory.AddConsole();
+
+            if (env.IsDevelopment())
+            {
+                app.UseDeveloperExceptionPage();
+            }
+
+            app.UseCors(policy => policy.WithOrigins("http://origin.sample.com:8080").WithMethods("GET"));
```
Eilon: Always use `*.example.com/org/net` as fake hostnames because they are reserved for such use: http://www.iana.org/domains/reserved

SampleDestination/project.json
```diff
@@ -0,0 +1,50 @@
...
+    "Microsoft.AspNetCore.Server.IISIntegration": "1.2.0-*",
+    "Microsoft.AspNetCore.Server.Kestrel": "1.2.0-*",
+    "Microsoft.Extensions.Logging.Console": "1.2.0-*",
+    "Microsoft.AspNetCore.Cors": "1.2.0-*",
+    "Microsoft.Extensions.Configuration.FileExtensions": "1.2.0-*",
+    "Microsoft.Extensions.Configuration.Json": "1.2.0-preview1-*"
```
Eilon: This one has preview1 in the version but the others don't?

SampleOrigin/wwwroot/Index.html
Eilon: Maybe keep it "somewhat" static but have 3 - 4 common scenarios? Or if it's easy enough to make it dynamic, that's fine too: just have textboxes with default values and a "Make request" button that uses the values. That would be neat.

test/Microsoft.AspNetCore.Cors.Test/CorsServiceTests.cs
```diff
@@ -227,6 +228,81 @@ public void EvaluatePolicy_PreflightRequest_MethodAllowed_ReturnsAllowMethods()
...
+                return new TheoryData<string, string, string[], string>
+                {
+                    {
+                        "http://example.com",
+                        "PUT",
+                        null,
+                        "Policy execution failed. Request origin http://example.com does not have permission to access the resource."
+                    },
+                    {
+                        "http://allowedexample.com",
```
Eilon: Also change these hosts. Can use sub-hosts if needed, e.g. allowed.example.com.

SampleOrigin/project.json
jbagga: I need it to launch the Index.html page in my origin to send CORS requests when I run SampleOrigin. Is there another way to launch the html page?

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
```diff
@@ -82,24 +97,41 @@ public CorsResult EvaluatePolicy(HttpContext context, CorsPolicy policy)
...
-            if (StringValues.IsNullOrEmpty(origin) || !policy.AllowAnyOrigin && !policy.Origins.Contains(origin))
+            if (StringValues.IsNullOrEmpty(origin))
+            {
+                return;
+            }
+
+            if (!policy.AllowAnyOrigin && !policy.Origins.Contains(origin))
             {
+                _logger?.RequestHasOriginHeader();
+                _logger.PolicyFailure($"Request origin {origin} does not have permission to access the resource.");
```
halter73: `?.`

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
jbagga: @Eilon, any suggestions for the verbosity/details for the log messages?

jbagga:
@kichalla @Tratcher @Eilon Updated. Working on changing the UI to accept custom methods

samples/README.md
```diff
@@ -0,0 +1,28 @@
...
+In the hosts file, add the following to the bottom of the file:
+127.0.0.1	destination.example.com
+127.0.0.1	origin.example.com
+
+Save the file and close it. Then clear your browser history.
+
+Run the sample
+*In a command prompt window, open the directory where you cloned the repository, and open the SampleDestination directory. Run the command: dotnet run
+*Repeat the above step in the SampleOrigin directory.
+*Open a browser window and go to http://origin.example.com
```
rynowak: Is it `http://origin.example.com:8080`?

rynowak: It would be good to add here the info about which port each of the apps will use by default

samples/SampleDestination/Startup.cs
```diff
@@ -0,0 +1,29 @@
...
+            services.AddCors();
+        }
+
+        public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory)
+        {
+            loggerFactory.AddConsole();
+            app.UseCors(policy => policy.WithOrigins("http://origin.example.com:8080"));
+            app.Run(async context =>
+            {
+                await context.Response.WriteAsync("Status code of your request: " + context.Response.StatusCode.ToString());
```
rynowak: Consider including here all of the cors-related headers

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
```diff
@@ -82,24 +97,43 @@ public CorsResult EvaluatePolicy(HttpContext context, CorsPolicy policy)
...
+            if (StringValues.IsNullOrEmpty(origin))
             {
+                _logger?.RequestDoesNotHaveOriginHeader();
                 return;
             }
 
+            if (!policy.AllowAnyOrigin && !policy.Origins.Contains(origin))
+            {
+                _logger?.RequestHasOriginHeader();
+                _logger?.PolicyFailure($"Request origin {origin} does not have permission to access the resource.");
```
rynowak: Do not use format strings in logging. Logging has it's own feature for this that's better

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
```diff
@@ -82,24 +97,43 @@ public CorsResult EvaluatePolicy(HttpContext context, CorsPolicy policy)
...
+                return;
+            }
+
+            _logger?.RequestHasOriginHeader();
```
rynowak: You could move this up to L105 and then you don't have to duplicate it

rynowak: You should also include the value of the origin header

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
```diff
@@ -124,7 +158,8 @@ public virtual void EvaluatePreflightRequest(HttpContext context, CorsPolicy pol
 
                 if (!found)
                 {
-                    return;
+                    _logger?.PolicyFailure($"Request method {accessControlRequestMethod} not allowed in CORS policy.");
```
rynowak: I think it makes sense to give each of these policy failures it's own event id and all the ceremony that comes with that. You probably also want a single "policy failed" message that doesn't provide any detail.

src/Microsoft.AspNetCore.Cors/Internal/CORSLoggerExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+        private static readonly Action<ILogger, Exception> _requestDoesNotHaveOriginHeader;
+        private static readonly Action<ILogger, Exception> _policySuccess;
+        private static readonly Action<ILogger, string, Exception> _policyFailure;
+
+        static CORSLoggerExtensions()
+        {
+            _isPreflightRequest = LoggerMessage.Define(
+                LogLevel.Debug,
+                1,
+                "This is a preflight request.");
```
rynowak: `The request is a preflight request.`

samples/SampleDestination/Program.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            var host = new WebHostBuilder()
+                .UseKestrel()
+                .UseUrls("http://*:5000")
+                .UseContentRoot(Directory.GetCurrentDirectory())
+                .UseIISIntegration()
```
Eilon: Is the sample Kestrel-only? If so, can probably remove this line?

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
```diff
@@ -133,6 +168,7 @@ public virtual void EvaluatePreflightRequest(HttpContext context, CorsPolicy pol
                 !requestHeaders.All(header => CorsConstants.SimpleRequestHeaders.Contains(header, StringComparer.OrdinalIgnoreCase) ||
```
jbagga: Does it make sense to refactor this a little, so we can log exactly which request header is not supported by the CORS policy? Maybe a foreach? @rynowak @kichalla

samples/SampleDestination/web.config
```diff
@@ -0,0 +1,14 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
Eilon: Ditto here regarding IIS - should be able to delete this.

samples/SampleOrigin/Program.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            var host = new WebHostBuilder()
+                .UseKestrel()
+                .UseUrls("http://*:8080")
+                .UseContentRoot(Directory.GetCurrentDirectory())
+                .UseIISIntegration()
```
Eilon: Delete

samples/SampleOrigin/Program.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace SampleOrigin
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            var host = new WebHostBuilder()
+                .UseKestrel()
+                .UseUrls("http://*:8080")
```
Eilon: From a sample perspective, it might be nice to use a pair of "matching" ports, e.g. 5000 and 5001, no?

samples/SampleOrigin/project.json
```diff
@@ -0,0 +1,46 @@
+{
+  "dependencies": {
+    "Microsoft.NETCore.App": {
+      "version": "1.1.0-*",
+      "type": "platform"
+    },
+    "Microsoft.AspNetCore.Server.IISIntegration": "1.2.0-*",
```
Eilon: Delete?

samples/SampleOrigin/project.json
```diff
@@ -0,0 +1,46 @@
...
+    "Microsoft.AspNetCore.Server.Kestrel": "1.2.0-*",
+    "Microsoft.Extensions.Logging.Console": "1.2.0-*"
+  },
+
+  "tools": {
+    "Microsoft.AspNetCore.Server.IISIntegration.Tools": "1.0.0-*"
```
Eilon: Delete.

samples/SampleOrigin/project.json
```diff
@@ -0,0 +1,46 @@
...
+
+  "publishOptions": {
+    "include": [
+      "wwwroot",
+      "web.config"
+    ]
+  },
+
+  "scripts": {
+    "postpublish": [ "dotnet publish-iis --publish-folder %publish:OutputPath% --framework %publish:FullTargetFramework%" ]
```
Eilon: Delete... ok basically delete anything that says "iis" in it 😄

samples/SampleOrigin/wwwroot/Index.html
```diff
@@ -0,0 +1,49 @@
...
+            };
+
+            request.send();
+        }
+
+    </script>
+    <title></title>
+</head>
+<body>
+    <form method="post">
```
Eilon: Probably don't need `<form>` tags at all, considering that:

1. They never get submitted
2. They both say `post` even though that doesn't reflect what they each do

(It's fine to have buttons outside of forms in HTML.... I'm pretty sure...)

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
```diff
@@ -133,6 +168,7 @@ public virtual void EvaluatePreflightRequest(HttpContext context, CorsPolicy pol
...
                                               policy.Headers.Contains(header, StringComparer.OrdinalIgnoreCase)))
             {
+                _logger?.PolicyFailure($"One or more request header(s) not allowed in CORS policy.");
```
Eilon: Also remove `$` from this string (it's not used, and it *shouldn't* be used).

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
kichalla: Yeah, that would be nice to know. This code path seems to be only for pre-flight request.

jbagga:
@Eilon @rynowak @kichalla Working on UI to define CORS policy (dynamically) in SampleDestination.

samples/README.md
```diff
@@ -0,0 +1,34 @@
...
+Both have different domain names, to simulate a CORS request. 
+
+Modify Hosts File
+Windows:
+Run a text editor (e.g. Notepad) as an Administrator. Open the hosts file on the path: "C:\Windows\System32\drivers\etc\hosts".
+
+Linux:
+On a Terminal window, type "sudo nano /etc/hosts" and enter your admin password when prompted.
+
+In the hosts file, add the following to the bottom of the file:
```
Eilon: Wrap the next two lines in triple-back-ticks.

Eilon: (Otherwise the Markdown doesn't render is nicely)

samples/README.md
```diff
@@ -0,0 +1,34 @@
...
+127.0.0.1	destination.example.com
+127.0.0.1	origin.example.com
+
+Save the file and close it. Then clear your browser history.
+
+Run the sample
+*In a command prompt window, open the directory where you cloned the repository, and open the SampleDestination directory. Run the command: dotnet run
+*Repeat the above step in the SampleOrigin directory.
+*Open a browser window and go to http://origin.example.com:5001
```
Eilon: Recommend putting things like URLs in back-tick "code" markers.

samples/SampleDestination/Startup.cs
```diff
@@ -0,0 +1,41 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Cors.Infrastructure;
```
Eilon: Is this namespace needed? Samples presumably shouldn't be using scary-sounding namespaces.

samples/SampleDestination/Startup.cs
```diff
@@ -0,0 +1,41 @@
...
+                .WithOrigins("http://origin.example.com:5001")              
+                .WithMethods("PUT")
+                .WithHeaders("Cache-Control"));
+
+            app.Run(async context =>
+            {
+                var responseHeaders = context.Response.Headers;
+                foreach (var responseHeader in responseHeaders)
+                {
+                    await context.Response.WriteAsync("\n"+responseHeader.Key+": "+responseHeader.Value);
```
Eilon: Formatting.

samples/SampleOrigin/Program.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace SampleOrigin
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            var host = new WebHostBuilder()
+                .UseKestrel()
+                .UseUrls("http://*:5001")
+                .UseContentRoot(Directory.GetCurrentDirectory())
```
Eilon: Hmm, it's actually not exactly right to use `GetCurrentDirectory()` because the current directory might not be the directory that contains the EXE. E.g. if I run the app by doing `..\another_folder\foo.exe` then the "current directory" is where "I am at" and not "where the executable is at".

Is there a better way to find the "directory we want"?

samples/SampleDestination/Startup.cs
```diff
@@ -0,0 +1,41 @@
...
+            loggerFactory.AddConsole();
+
+            app.UseCors(policy => policy               
+                .WithOrigins("http://origin.example.com:5001")              
+                .WithMethods("PUT")
+                .WithHeaders("Cache-Control"));
+
+            app.Run(async context =>
+            {
+                var responseHeaders = context.Response.Headers;
```
Tratcher: ContentType = "text/plain"

samples/SampleDestination/Startup.cs
Tratcher: spacing

samples/SampleOrigin/web.config
```diff
@@ -0,0 +1,9 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <system.webServer>
+    <handlers>
```
Eilon: Delete this file? (IIS-only?)

samples/SampleOrigin/Startup.cs
```diff
@@ -0,0 +1,36 @@
...
+            {
+                var fileInfoProvider = env.WebRootFileProvider;
+                var fileInfo = fileInfoProvider.GetFileInfo("/Index.html");
+                context.Response.Headers.Add("Content-Type", "text/html; charset=utf-8");
+                return context.Response.SendFileAsync(fileInfo);
+            });
+
+            app.Run(async context =>
+            {
+                await context.Response.WriteAsync("Status code of your request: " + context.Response.StatusCode.ToString());
```
Tratcher: This code will never run, all requests are handled by the prior Run.

samples/SampleOrigin/wwwroot/Index.html
```diff
@@ -0,0 +1,99 @@
...
+            };
+
+            request.send();
+        }
+    </script>
+    <p>CORS Sample</p>
+    Method: <input type="text" id="methodName" /><br /><br />
+    Header Name: <input type="text" id="headerName" value="Cache-Control" /> Header Value: <input type="text" id="headerValue" value="no-cache" /><br /><br />
+    <script>
+        document.getElementById('methodName')
```
Eilon: Is the JS really needed? Can the user just tab between the fields?

test/Microsoft.AspNetCore.Cors.Test/CorsServiceTests.cs
```diff
@@ -913,5 +1070,15 @@ public void ApplyResult_PreflightMaxAge_MaxAgeHeaderAdded()
 
             return context;
         }
+
+        public struct LogData
+        {
+            public string origin { get; set; }
```
Eilon: Properties should be PascalCased?

test/Microsoft.AspNetCore.Cors.Test/CorsServiceTests.cs
```diff
@@ -913,5 +1070,15 @@ public void ApplyResult_PreflightMaxAge_MaxAgeHeaderAdded()
 
             return context;
         }
+
+        public struct LogData
```
Eilon: Why `struct` and not `class`?

samples/README.md
```diff
@@ -0,0 +1,34 @@
+CORS Sample
+===
+This sample consists of a request origin (SampleOrigin) and a request destination (SampleDestination). 
+Both have different domain names, to simulate a CORS request. 
+
+Modify Hosts File
+Windows:
```
Eilon: This Markdown also won't render quite right: Markdown generally ignores single newlines. Recommend adding some additional formatting, such as bullet-points, section headings, etc. to get it to format correctly.

This section should probably also say *why* you have to modify the hosts file. E.g. "To run this CORS sample, modify the `hosts` file to register the hostnames:"

samples/SampleOrigin/Program.cs
Tratcher: GetCurrentDirectory is what we use in all the samples and templates.

Eilon: For real? Isn't that totally wrong? Or does dotnet.exe change the current directory? (Because if it does, that's an even worse bug!)

Tratcher: @davidfowl

Eilon: (It's not some super critical issue, so I'm certainly not blocking this PR on that, but we should resolve that question - I'll start a mail thread.)

jbagga:
@Eilon @rynowak Updated

samples/SampleOrigin/Program.cs
rynowak: Uhhhhh

rynowak: @Eilon - this has been "the way to do it" since we dropped DNX.

Eilon: Yeah per the mail thread this might actually be *necessary* for some scenarios. But I still content it isn't exactly right. But let's leave it.

samples/SampleOrigin/Startup.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        public void ConfigureServices(IServiceCollection services)
+        {
+        }
+
+        public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory)
+        {
+            loggerFactory.AddConsole();
+            app.Run(context =>
+           {
```
Eilon: Wrong indentation!

samples/SampleOrigin/wwwroot/Index.html
```diff
@@ -0,0 +1,88 @@
...
+            request.send();
+        }
+    </script>
+
+    <p>CORS Sample</p>
+    Method: <input type="text" id="methodName" /><br /><br />
+    Header Name: <input type="text" id="headerName" /> Header Value: <input type="text" id="headerValue" /><br /><br />
+    <script>
+        document.getElementById('headerValue')
+    .addEventListener("keyup", function (event) {
```
Eilon: Indentation!

jbagga:
@Eilon Updated

samples/README.md
```diff
@@ -0,0 +1,38 @@
+# CORS Sample
+
+This sample consists of a request origin (SampleOrigin) and a request destination (SampleDestination). Both have different domain names, to simulate a CORS request. 
+
+## Modify Hosts File
+To run this CORS sample, modify the hosts file to register the hostnames `destination.example.com` and `origin.example.com.`
```
Eilon: The last `.` in `origin.example.com.` should be outside the `

samples/SampleDestination/Startup.cs
davidfowl: We don't set the content type in any other sample I'm aware of...

PR #106, opened 2017-03-20
Inject ICorsPolicyProvider instance through Invoke

### LINKED ISSUE #105: CorsMiddleware should resolve dependencies dynamically
The dependency on `ICorsService` and `ICorsPolicyProvider` to the [ctor for the middleware](https://github.com/aspnet/CORS/blob/dev/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsMiddleware.cs#L31-L32) means the use of these services are in effect singletons. This means implementations that require scoping in DI (e.g. to hit a database) will not work properly and the one instance will be used concurrently. I'd suggest a change in `Invoke` to resolve these dynamically.

henkmollema:
This allows for scoped instances of `ICorsPolicyProvider` to be injected in the CORS middleware and prevents turning them into singletons.

Resolves #105

### DIFF
--- a/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsMiddleware.cs
@@ -15,7 +15,6 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
     {
         private readonly RequestDelegate _next;
         private readonly ICorsService _corsService;
-        private readonly ICorsPolicyProvider _corsPolicyProvider;
         private readonly CorsPolicy _policy;
         private readonly string _corsPolicyName;
 
@@ -24,12 +23,10 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
         /// </summary>
         /// <param name="next">The next middleware in the pipeline.</param>
         /// <param name="corsService">An instance of <see cref="ICorsService"/>.</param>
-        /// <param name="policyProvider">A policy provider which can get an <see cref="CorsPolicy"/>.</param>
         /// <param name="policyName">An optional name of the policy to be fetched.</param>
         public CorsMiddleware(
             RequestDelegate next,
             ICorsService corsService,
-            ICorsPolicyProvider policyProvider,
             string policyName)
         {
             if (next == null)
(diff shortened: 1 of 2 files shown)


henkmollema:
@Eilon are we going forward with this PR? And if so, do you want me to update (rebase) it?

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsMiddleware.cs
```diff
@@ -85,11 +76,11 @@ public class CorsMiddleware
         }
 
         /// <inheritdoc />
-        public async Task Invoke(HttpContext context)
+        public async Task Invoke(HttpContext context, ICorsPolicyProvider corsPolicyProvider)
         {
             if (context.Request.Headers.ContainsKey(CorsConstants.Origin))
             {
-                var corsPolicy = _policy ?? await _corsPolicyProvider?.GetPolicyAsync(context, _corsPolicyName);
+                var corsPolicy = _policy ?? await corsPolicyProvider?.GetPolicyAsync(context, _corsPolicyName);
```
davidfowl: Pretty sure this can't be null.

henkmollema: You're right. Invoking middleware will fail if it can't resolve the parameters, right? I'll remove the `?`.

henkmollema:
@davidfowl @Eilon updated and rebased.

henkmollema:
We need to update baseline.json as well.

Eilon:
BTW per the discussion in https://github.com/aspnet/CORS/issues/105, this is a breaking change, so it can't be taken as-is in the 2.1 release, so this PR is "on hold" until we figure out how to resolve that.

aspnet-hello:
This issue was moved to aspnet/Home#2376

aspnet-hello:
Sorry I was a bad bot. Re-opening this PR.

Eilon:
@mkArtakMSFT  - any further action planned on this PR?

I think that given it's a breaking change, I recommend we close it.

pranavkm:
@Eilon close?

brockallen:
Why close? It needs to be fixed at some point.

brockallen:
> It's likely a breaking change would necessitate waiting until 3.0.

Sure. Don't you have a way of assigning/labeling the PR for 3.0?

mkArtakMSFT:
Resolving the moved issue: https://github.com/aspnet/Home/issues/2377

mkArtakMSFT:
@henkmollema, we do want to take this change in, and we've just marked the referenced issue accordingly. However, this PR is outdated already and given that we aren't going to act on it yet, you can close it if you want and we'll do the change when the right time will come.

pranavkm:
Tracking merging this in #202

PR #126, opened 2017-09-09
Updated EnableCorsAttribute and UseCors extension to not have to supp…

### LINKED ISSUE #122: Add an empty constructor for EnableCorsAttribute
To use the default CORS policy, we need to pass in either `null` or `CorsOptions.DefaultPolicyName`. The latter is an instance member and therefore cannot be used in an attribute. It would be nice to substitute  `[EnableCors(null)]` with `[EnableCors]`. Additionally, it would be nice if we had an (extension?) method on `CorsOptions` that allowed setting the default policy e.g. `options.AddDefaultCorsPolicy(builder => ..` instead of doing `options.AddPolicy(options.DefaultPolicyName, builder => `

kichalla:
…ly a policy name.
[Fixes #122] Add an empty constructor for EnableCorsAttribute

### DIFF
--- a/src/Microsoft.AspNetCore.Cors/EnableCorsAttribute.cs
+++ b/src/Microsoft.AspNetCore.Cors/EnableCorsAttribute.cs
@@ -11,7 +11,16 @@ namespace Microsoft.AspNetCore.Cors
     public class EnableCorsAttribute : Attribute, IEnableCorsAttribute
     {
         /// <summary>
-        /// Creates a new instance of the <see cref="EnableCorsAttribute"/>.
+        /// Creates a new instance of the <see cref="EnableCorsAttribute"/> with the default policy
+        /// name defined by <see cref="CorsOptions.DefaultPolicyName"/>.
+        /// </summary>
+        public EnableCorsAttribute()
+            : this(policyName: null)
+        {
+        }
+
+        /// <summary>
+        /// Creates a new instance of the <see cref="EnableCorsAttribute"/> with the supplied policy name.
         /// </summary>
         /// <param name="policyName">The name of the policy to be applied.</param>
         public EnableCorsAttribute(string policyName)
--- a/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsOptions.cs
+++ b/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsOptions.cs
@@ -32,6 +32,34 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
         }
 
         /// <summary>
+        /// Adds a new policy and sets it as the default.
+        /// </summary>
+        /// <param name="policy">The <see cref="CorsPolicy"/> policy to be added.</param>
+        public void AddDefaultPolicy(CorsPolicy policy)
+        {
+            if (policy == null)
+            {
+                throw new ArgumentNullException(nameof(policy));
+            }
+
+            AddPolicy(DefaultPolicyName, policy);
+        }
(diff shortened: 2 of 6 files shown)


src/Microsoft.AspNetCore.Cors/EnableCorsAttribute.cs
```diff
@@ -13,6 +13,14 @@ public class EnableCorsAttribute : Attribute, IEnableCorsAttribute
         /// <summary>
         /// Creates a new instance of the <see cref="EnableCorsAttribute"/>.
```
pranavkm: with the default policy

pranavkm: Let's call this out in the other overloads that were added.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsOptions.cs
```diff
@@ -32,6 +32,48 @@ public string DefaultPolicyName
         }
 
         /// <summary>
+        /// Adds a new policy and sets it as the default.
+        /// </summary>
+        /// <param name="name">The name of the policy.</param>
+        /// <param name="policy">The <see cref="CorsPolicy"/> policy to be added.</param>
+        public void AddDefaultPolicy(string name, CorsPolicy policy)
```
pranavkm: Wouldn't we just use `DefaultPolicyName` here? Essentially have this method signature be `AddDefaultPolicy(CorsPolicy policy)` which internally calls `AddPolicy(DefaultPolicy, policy)`

kichalla: But this would require 2 lines of code (like setting the DefaultPolicyName and then doing AddDefaultPolicy(CorsPolicy)) and it implies an order too

pranavkm: Don't we assign it a default name anyway? I'm not super familiar with these pieces, but I'd imagine a code path where I don't care about the name at all:

```
AddDefaultPolicy(new CorsPolicy { .. })

[EnableCore]
public IActionResult Foo()
```

^ No names involved.

Eilon: Yeah I don't think you need the user to pass in a name here... unless I'm also missing something.

Eilon: Though I admit it's odd to allow setting the default policy name. The default *should* have been some magical readonly token (or empty string or something).

Eilon (changes requested):

test/Microsoft.AspNetCore.Cors.Test/CorsOptionsTest.cs
```diff
@@ -0,0 +1,67 @@
...
+                expectedPolicy = policyBuilder.AllowAnyOrigin().Build();
+            });
+            corsOptions.AddPolicy("policy3", new CorsPolicy());
+
+            // Assert
+            var actualPolicy = corsOptions.GetPolicy(corsOptions.DefaultPolicyName);
+            Assert.Same(expectedPolicy, actualPolicy);
+        }
+    }
+}
```
pranavkm: new line at EOF

pranavkm (approved):

Eilon (approved):

PR #139, opened 2018-01-04
Handle malformed origin Uri

jbagga:
Addresses https://github.com/aspnet/Home/issues/2318

cc @Tratcher

### DIFF
--- a/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyExtensions.cs
+++ b/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyExtensions.cs
@@ -17,19 +17,15 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
                 return true;
             }
 
-            Uri.TryCreate(origin, UriKind.Absolute, out var originUri);
-
-            if (originUri != null)
+            if (Uri.TryCreate(origin, UriKind.Absolute, out var originUri))
             {
                 return policy.Origins
                     .Where(o => o.Contains($"://{_WildcardSubdomain}"))
                     .Select(CreateDomainUri)
                     .Any(domain => UriHelpers.IsSubdomainOf(originUri, domain));
             }
-            else
-            {
-                return false;
-            }
+
+            return false;
         }
 
         private static Uri CreateDomainUri(string origin)
--- a/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyExtensions.cs
+++ b/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyExtensions.cs
@@ -16,7 +16,9 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
             {
                 return true;
             }
-            var originUri = new Uri(origin, UriKind.Absolute);
+
+            Uri.TryCreate(origin, UriKind.Absolute, out var originUri);
+
             return policy.Origins
                 .Where(o => o.Contains($"://{_WildcardSubdomain}"))
                 .Select(CreateDomainUri)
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyExtensions.cs
```diff
@@ -16,7 +16,9 @@ public static bool IsOriginAnAllowedSubdomain(this CorsPolicy policy, string ori
             {
                 return true;
             }
-            var originUri = new Uri(origin, UriKind.Absolute);
+
+            Uri.TryCreate(origin, UriKind.Absolute, out var originUri);
```
Tratcher: And if it fails? Does UriHelpers.IsSubdomainOf accept null?

jbagga: Yeah but it returns false here https://github.com/aspnet/CORS/blob/dev/src/Microsoft.AspNetCore.Cors/Infrastructure/UriHelpers.cs#L12

Tratcher: https://github.com/aspnet/CORS/blob/c873fc55f9fe8ff5d0de2e54cd763cd4a776c42b/src/Microsoft.AspNetCore.Cors/Infrastructure/UriHelpers.cs#L13 will null ref if you ever get there.

jbagga: Oh my tests were short circuiting before even going there. Let me look into it. Thanks

jbagga:
@Tratcher Updated

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyExtensions.cs
```diff
@@ -16,11 +16,20 @@ public static bool IsOriginAnAllowedSubdomain(this CorsPolicy policy, string ori
...
             }
-            var originUri = new Uri(origin, UriKind.Absolute);
-            return policy.Origins
-                .Where(o => o.Contains($"://{_WildcardSubdomain}"))
-                .Select(CreateDomainUri)
-                .Any(domain => UriHelpers.IsSubdomainOf(originUri, domain));
+
+            Uri.TryCreate(origin, UriKind.Absolute, out var originUri);
+
+            if (originUri != null)
```
Tratcher: Use Uri.TryCreate as the if condition.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyExtensions.cs
```diff
@@ -16,11 +16,20 @@ public static bool IsOriginAnAllowedSubdomain(this CorsPolicy policy, string ori
...
+            {
+                return policy.Origins
+                    .Where(o => o.Contains($"://{_WildcardSubdomain}"))
+                    .Select(CreateDomainUri)
+                    .Any(domain => UriHelpers.IsSubdomainOf(originUri, domain));
+            }
+            else
```
Tratcher: flatten else, the if returned.

test/Microsoft.AspNetCore.Cors.Test/CorsPolicyExtensionsTests.cs
```diff
@@ -23,6 +23,25 @@ public void IsOriginAnAllowedSubdomain_ReturnsTrueIfPolicyContainsOrigin()
             Assert.True(actual);
         }
 
+        [Theory]
+        [InlineData(null)]
+        [InlineData("null")]
+        [InlineData("http://")]
+        [InlineData("http://*")]
+        [InlineData("http://.domain")]
+        public void IsOriginAnAllowedSubdomain_ReturnsFalseIfOriginIsMalformedUri(string malformedOrigin)
```
rynowak: Should we also test some URLs that have something after the authority portion? like: `http://foo.bar/kdkfdkfdkf`

jbagga: This test is specifically for malformed origins. Other tests cover well-formed Uris. I have added a test for the following: `"http://.domain/hello")]`

jbagga:
@rynowak Updated

PR #142, opened 2018-02-02
Fix duplicate header names

priyanshu92:
This uses http header and method names from strings defined in [HeaderNames](https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.Net.Http.Headers/HeaderNames.cs) and [HttpMethods](https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Http.Abstractions/HttpMethods.cs) respectively.

Resolves aspnet/home#2319


priyanshu92:
@mkArtakMSFT  Yeah sure. Will push a new commit.

PR #154, opened 2018-03-07
Prkrishn/test1

pranavkm:


build/buildpipeline/linux.groovy
```diff
@@ -5,6 +5,11 @@ simpleNode('Ubuntu16.04', 'latest-or-auto-docker') {
         checkout scm
     }
     stage ('Build') {
+        environment {
```
pranavkm: @mmitche here's a dumb question - how do I set env variables? I tried withEnv and that didn't work either. It works if I specify it in parameters, but that makes the jobName (and consequently the build logs) less pretty.

PR #163, opened 2018-06-12
No CORS headers sent if Exception is thrown

Lavinski:
Normally headers are added however if a controller throws an exception then CORS headers not be present.

All matching responses should have CORS headers added. This problem affects the use of middleware (including the UseExceptionHandler middleware).

This PR fixes the issue by moving the adding of the headers to `Response.OnStarting`.

https://github.com/aspnet/Home/issues/2378
https://github.com/aspnet/Home/issues/3220

### DIFF
--- a/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsMiddleware.cs
@@ -106,26 +106,38 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
                 var corsPolicy = _policy ?? await _corsPolicyProvider?.GetPolicyAsync(context, _corsPolicyName);
                 if (corsPolicy != null)
                 {
-                    var corsResult = _corsService.EvaluatePolicy(context, corsPolicy);
-                    _corsService.ApplyResult(corsResult, context.Response);
-
                     var accessControlRequestMethod =
                         context.Request.Headers[CorsConstants.AccessControlRequestMethod];
                     if (string.Equals(
                             context.Request.Method,
                             CorsConstants.PreflightHttpMethod,
                             StringComparison.OrdinalIgnoreCase) &&
-                            !StringValues.IsNullOrEmpty(accessControlRequestMethod))
+                        !StringValues.IsNullOrEmpty(accessControlRequestMethod))
                     {
+                        ApplyCorsHeaders(context, corsPolicy);
+
                         // Since there is a policy which was identified,
                         // always respond to preflight requests.
                         context.Response.StatusCode = StatusCodes.Status204NoContent;
                         return;
                     }
+                    else
+                    {
+                        context.Response.OnStarting(_ => {
+                            ApplyCorsHeaders(context, corsPolicy);
+                            return Task.CompletedTask;
+                        }, context);
+                    }
                 }
             }
 
             await _next(context);
+        }
+
+        private void ApplyCorsHeaders(HttpContext context, CorsPolicy corsPolicy)
+        {
+            var corsResult = _corsService.EvaluatePolicy(context, corsPolicy);
+            _corsService.ApplyResult(corsResult, context.Response);
         }
     }
 }
--- a/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsMiddleware.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.AspNetCore.Cors.Infrastructure
@@ -18,6 +19,7 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
         private readonly ICorsPolicyProvider _corsPolicyProvider;
         private readonly CorsPolicy _policy;
         private readonly string _corsPolicyName;
+        private readonly ILogger _logger;
 
         /// <summary>
         /// Instantiates a new <see cref="CorsMiddleware"/>.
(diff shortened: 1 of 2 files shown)


Lavinski:
This feature is ready from a development position. I don't foresee any breaking changes and this is a highly requested feature. Ready for review.

rynowak:
@Tratcher - do you have thoughts about this? I'm deeply passionate about us fixing this issue, we get this feedback a lot.

The change itself is designed to make an end-run around `HttpResponse.Clear()` - I'm not aware of anywhere else we do this. Funnily enough to the exception handling middleware sets cache heads to prevent caching on failures (developer exception page does not). Other there other things that need to work the same way as CORS (CSP for example) or is CORS special?

rynowak:
Note that this is also duplicated in MVC - and it looks like it would need the same changes there. https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Cors/CorsAuthorizationFilter.cs

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsMiddleware.cs
```diff
@@ -106,26 +106,38 @@ public class CorsMiddleware
...
+
                         // Since there is a policy which was identified,
                         // always respond to preflight requests.
                         context.Response.StatusCode = StatusCodes.Status204NoContent;
                         return;
                     }
+                    else
+                    {
+                        context.Response.OnStarting(_ => {
+                            ApplyCorsHeaders(context, corsPolicy);
```
Tratcher: Catch all exceptions in this callback. An exception on this call path would bypass all exception pages and result in a blank 500.

Lavinski: This is middleware which means it can be used without mvc. Furthermore exceptions on this path already result in a blank 500 except it won’t have CORS headers.

You should use the exception handler middleware to provide consistent, last chance errors to the client. This will also help in cases were mvc is used with other serving middleware.

Tratcher: It's not practical for the exception handling middleware to handle exceptions in OnStarting, those callbacks are invoked directly by the server.

Lavinski: I’m sorry I think you’ve misunderstood (or I’m missing something). There is no exception handling in OnStarting. 

This change is primarily to ensure CORS middleware works with ‘UseExceptionHandler’ but applies to all similar cases.

The flow goes like this:
1. An exception is uncaught in an API controller.
2. Exception is handled by global exception handler.
3. CORS headers are added to error details can be used by the client.

Is there anything else I can clarify?

JulijaRamoskiene:
I think it is a mistake to have duplicated code in CorsAuthorizationFilter and CorsMiddleware E.g.: preflight requests are now handled differently between them. CorsAuthorizationFilter sets 200 OK response code and CorsMiddleware sets 204 No Content.

JulijaRamoskiene:
Since CorsAuthorizationFilter and CorsMiddleware both call ApplyResult () on ICorsService, maybe it would make sense to move the fix to that method?

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsMiddleware.cs
Lavinski: @Tratcher or did you mean wrapping `ApplyCorsHeaders` in a `try catch` in case of an error in the callback?
```
context.Response.OnStarting(_ => {
    ApplyCorsHeaders(context, corsPolicy);
```

Tratcher: Right. We want to avoid throwing exceptions in the OnStarting code path as it has a worse error handling experience. These wouldn't be app exceptions, but exceptions from within CORS itself. I'm not aware of any, but we'd rather not risk it.

Lavinski:
@Tratcher I've added a `try` for code inside `OnStarting` and I've changed the `OnStarting` to pass through the context (as opposed to using the variable via a closure).

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsMiddleware.cs
```diff
@@ -18,18 +19,21 @@ public class CorsMiddleware
...
         /// <param name="corsService">An instance of <see cref="ICorsService"/>.</param>
         /// <param name="policyProvider">A policy provider which can get an <see cref="CorsPolicy"/>.</param>
+        /// <param name="loggerFactory">An instance of <see cref="ILoggerFactory"/>.</param>
         public CorsMiddleware(
             RequestDelegate next,
             ICorsService corsService,
-            ICorsPolicyProvider policyProvider)
-            : this(next, corsService, policyProvider, policyName: null)
+            ICorsPolicyProvider policyProvider,
+            ILoggerFactory loggerFactory)
```
Tratcher: AppVeyor is warning about API signature breaks here.
https://ci.appveyor.com/project/aspnetci/cors/build/1.0.476

For this 2.2 milestones we cannot change any existing APIs, only add new overloads.

Tratcher: See what I did over here:
https://github.com/aspnet/BasicMiddleware/pull/340/files

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsMiddleware.cs
```diff
@@ -106,26 +125,45 @@ public class CorsMiddleware
...
                         // always respond to preflight requests.
                         context.Response.StatusCode = StatusCodes.Status204NoContent;
                         return;
                     }
+                    else
+                    {
+                        context.Response.OnStarting(c => {
+                            try
+                            {
+                                ApplyCorsHeaders((HttpContext)c, corsPolicy);
```
Tratcher: Pass both parameters in via state like this: https://github.com/aspnet/KestrelHttpServer/pull/2670/files#diff-4ab77515238e5e964ae9a3a4615205b1R713
Then this callback can become a cached instance.

pranavkm:
Merged via https://github.com/aspnet/CORS/commit/554855cab34961a27a6cf248fbb847b9dd8bd8d4. Thanks for the PR!

PR #166, opened 2018-06-25
No CORS headers sent if Exception is thrown

pranavkm:
- Normally headers are added however if a controller throws an exception then CORS headers not be present.

Addresses aspnet/Home#3220

### DIFF
--- a/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsMiddleware.cs
+++ b/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsMiddleware.cs
@@ -3,7 +3,10 @@
 
 using System;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Cors.Internal;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.AspNetCore.Cors.Infrastructure
@@ -18,6 +21,7 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
         private readonly ICorsPolicyProvider _corsPolicyProvider;
         private readonly CorsPolicy _policy;
         private readonly string _corsPolicyName;
+        private readonly ILogger _logger;
 
         /// <summary>
         /// Instantiates a new <see cref="CorsMiddleware"/>.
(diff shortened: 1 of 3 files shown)


pranavkm:
Cleaned up formatting + LoggerExtensions

src/Microsoft.AspNetCore.Cors/Infrastructure/LoggerExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    internal static class LoggerExtensions
+    {
+        private static readonly Action<ILogger, Exception> _failedToSetCorsHeaders;
+
+        static LoggerExtensions()
+        {
+            _failedToSetCorsHeaders = LoggerMessage.Define(
+                LogLevel.Warning,
+                1,
```
kichalla: There should be other logging events from Cors, this new one should be after them

test/Microsoft.AspNetCore.Cors.Test/CorsMiddlewareTests.cs
```diff
@@ -349,5 +354,39 @@ public class CorsMiddlewareTests
...
+        {
+            // Arrange
+            var hostBuilder = new WebHostBuilder()
+                .Configure(app => {
+                    app.UseCors(builder =>
+                        builder.WithOrigins("http://localhost:5001")
+                            .WithMethods("PUT")
+                            .WithHeaders("Header1")
+                            .WithExposedHeaders("AllowedHeader"));
+                    app.Run(async context => {
```
kichalla: Would be good to mimic a similar scenario as the original issue (i.e to register a middleware _before_ Cors and which catches exceptions globally and clears the response headers and sends out a generic safe response)

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsMiddleware.cs
```diff
@@ -40,10 +44,61 @@ public class CorsMiddleware
...
+        {
+        }
+
+        /// <summary>
+        /// Instantiates a new <see cref="CorsMiddleware"/>.
+        /// </summary>
+        /// <param name="next">The next middleware in the pipeline.</param>
+        /// <param name="corsService">An instance of <see cref="ICorsService"/>.</param>
+        /// <param name="policy">An instance of the <see cref="CorsPolicy"/> which can be applied.</param>
+        [Obsolete("This constructor has been replaced with an equivalent constructor which requires an ILoggerFactory")]
```
kichalla: Remove this constructor?

pranavkm: GitHub is doing a bad job doing a diff. There were 3 original ctos - https://github.com/aspnet/CORS/blob/dev/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsMiddleware.cs - that now have an overload each.

src/Microsoft.AspNetCore.Cors/Infrastructure/LoggerExtensions.cs
pranavkm: Ahh, didn't notice it was sitting under the Internal ns.

test/Microsoft.AspNetCore.Cors.Test/CorsMiddlewareTests.cs
```diff
@@ -349,5 +355,121 @@ public class CorsMiddlewareTests
                 Assert.Equal("PUT", response.Headers.GetValues(CorsConstants.AccessControlAllowMethods).FirstOrDefault());
             }
         }
+
+        [Fact]
+        public async Task CorsRequest_AfterResponseCleared_SetsResponseHeaders()
```
kichalla: nit: change name

PR #170, opened 2018-07-05
Lowercase scheme and host of origins to better align with RFC 6454.

sindhudweep:
Lowercase scheme and host of origins.
 - Many browsers (clients) typically follow RFC 6454 and lower case the scheme and host members of the origin triple. See https://tools.ietf.org/html/rfc6454 § 4.2 and 4.5

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -2,7 +2,7 @@ init:
 - git config --global core.autocrlf true
 branches:
   only:
-  - dev
+  - master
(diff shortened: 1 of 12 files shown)


pranavkm:
Merged as part of https://github.com/aspnet/CORS/commit/feceb3ddeaf0a0ca7e895eb83674d78fee53515c. Thanks for the PR!

PR #173, opened 2018-07-18
Fixed preflight simple request issue with content-type.

FlyingWraptor:
This merge fixes issue aspnet/Home/issues/3323

As explained in the issue I've found that the options method does not return any `Access-Control-Allow-Method` header for GET/POST or HEAD, though Angular requires it. Doing a bit more research I've found the following (requoted from issue comment) condition:
>      Or if the Content-Type header has a value other than the following:
>          application/x-www-form-urlencoded
>          multipart/form-data
>          text/plain
Source: [Mozilla dev documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Preflighted_requests)
Meaning if the client requires some other value than the specified above it does a preflight check using 'simple methods' `GET/POST or HEAD` but providing a `Access-Control-Request-Headers` header with content-type as value.

So this fix returns if provided the right criteria (Required header and 'simple method') skips the filtering of the 'Simple methods' and handles the request as prescribed in the [Fetch](https://fetch.spec.whatwg.org/) spec.

### DIFF
--- a/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsResult.cs
+++ b/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsResult.cs
@@ -45,6 +45,11 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
         public bool VaryByOrigin { get; set; }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


FlyingWraptor:
@Eilon

PR #174, opened 2018-07-18
Fixed preflighted simple request issue with content-type.

FlyingWraptor:
This merge fixes issue aspnet/Home/issues/3323

As explained in the issue I've found that the options method does not return any `Access-Control-Allow-Method` header for GET/POST or HEAD, though Angular requires it. Doing a bit more research I've found the following (requoted from issue comment) condition:
>      Or if the Content-Type header has a value other than the following:
>          application/x-www-form-urlencoded
>          multipart/form-data
>          text/plain
Source: [Mozilla dev documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Preflighted_requests)
Meaning if the client requires some other value than the specified above it does a preflight check using 'simple methods' `GET/POST or HEAD` but providing a `Access-Control-Request-Headers` header with content-type as value.

So this fix returns if provided the right criteria (Required header and 'simple method') skips the filtering of the 'Simple methods' and handles the request as prescribed in the [Fetch](https://fetch.spec.whatwg.org/) spec.

### DIFF
--- a/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
+++ b/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
@@ -151,8 +151,7 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
             {
                 foreach (var requestHeader in requestHeaders)
                 {
-                    if (!CorsConstants.SimpleRequestHeaders.Contains(requestHeader, StringComparer.OrdinalIgnoreCase) &&
-                                                  !policy.Headers.Contains(requestHeader, StringComparer.OrdinalIgnoreCase))
+                    if (!policy.Headers.Contains(requestHeader, StringComparer.OrdinalIgnoreCase))
                     {
                         _logger?.PolicyFailure();
                         _logger?.RequestHeaderNotAllowed(requestHeader);
@@ -201,50 +200,23 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
 
             if (result.AllowedMethods.Count > 0)
             {
-                // Filter out simple methods
-                var nonSimpleAllowMethods = result.AllowedMethods
-                    .Where(m =>
-                        !CorsConstants.SimpleMethods.Contains(m, StringComparer.OrdinalIgnoreCase))
-                    .ToArray();
-
-                if (nonSimpleAllowMethods.Length > 0)
-                {
-                    headers.SetCommaSeparatedValues(
-                        CorsConstants.AccessControlAllowMethods,
-                        nonSimpleAllowMethods);
-                }
+                headers.SetCommaSeparatedValues(
+                    CorsConstants.AccessControlAllowMethods,
+                    result.AllowedMethods.ToArray());
             }
 
             if (result.AllowedHeaders.Count > 0)
             {
-                // Filter out simple request headers
-                var nonSimpleAllowRequestHeaders = result.AllowedHeaders
-                    .Where(header =>
-                        !CorsConstants.SimpleRequestHeaders.Contains(header, StringComparer.OrdinalIgnoreCase))
-                    .ToArray();
-
-                if (nonSimpleAllowRequestHeaders.Length > 0)
-                {
-                    headers.SetCommaSeparatedValues(
-                        CorsConstants.AccessControlAllowHeaders,
-                        nonSimpleAllowRequestHeaders);
-                }
+                headers.SetCommaSeparatedValues(
+                    CorsConstants.AccessControlAllowHeaders,
+                    result.AllowedHeaders.ToArray());
             }
 
             if (result.AllowedExposedHeaders.Count > 0)
             {
-                // Filter out simple response headers
-                var nonSimpleAllowResponseHeaders = result.AllowedExposedHeaders
-                    .Where(header =>
-                        !CorsConstants.SimpleResponseHeaders.Contains(header, StringComparer.OrdinalIgnoreCase))
-                    .ToArray();
-
-                if (nonSimpleAllowResponseHeaders.Length > 0)
-                {
-                    headers.SetCommaSeparatedValues(
-                        CorsConstants.AccessControlExposeHeaders,
-                        nonSimpleAllowResponseHeaders);
-                }
+                headers.SetCommaSeparatedValues(
+                    CorsConstants.AccessControlExposeHeaders,
+                    result.AllowedExposedHeaders.ToArray());
             }
 
             if (result.PreflightMaxAge.HasValue)
--- a/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsConstants.cs
+++ b/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsConstants.cs
@@ -65,31 +65,5 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
         /// The Access-Control-Max-Age response header.
         /// </summary>
         public static readonly string AccessControlMaxAge = HeaderNames.AccessControlMaxAge;
-
-
-        internal static readonly string[] SimpleRequestHeaders =
-        {
-            HeaderNames.Origin,
-            HeaderNames.Accept,
-            HeaderNames.AcceptLanguage,
-            HeaderNames.ContentLanguage,
-        };
-
-        internal static readonly string[] SimpleResponseHeaders =
-        {
-            HeaderNames.CacheControl,
-            HeaderNames.ContentLanguage,
-            HeaderNames.ContentType,
-            HeaderNames.Expires,
-            HeaderNames.LastModified,
-            HeaderNames.Pragma
-        };
-
-        internal static readonly string[] SimpleMethods =
-        {
-            HttpMethods.Get,
-            HttpMethods.Head,
-            HttpMethods.Post
-        };
     }
 }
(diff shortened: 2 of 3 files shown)


FlyingWraptor:
@Eilon

Eilon:
@mkArtakMSFT - can you have someone take a look?

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsResult.cs
```diff
@@ -44,6 +44,11 @@ public class CorsResult
         /// </summary>
         public bool VaryByOrigin { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating if 'Simple methods' should be filtered.
```
Eilon: Even the Mozilla docs don't use the term "simple methods" (at most it says "simple requests"). But given that even those docs concede that they made the term up themselves, can we change the terminology here to be more specific and avoid made-up terms? (In both the docs and the API naming.)

FlyingWraptor: Simple requests use the 'Simple Methods' which are defined in already commited changes. I adopted the terminology from the CorsConstants.SimpleMethods variable defined already containing GET POST and HEAD HttpMethods. 

I found it to be suiting to the already in-place filter.

Eilon: Oh we already use that naming somewhere? I didn't realize that, my apologies! Perhaps this is fine then.

Eilon: Thanks @FlyingWraptor , I'll let the experts in this area provide further comments as to whether this is fitting. Based on what you've said, this does seem reasonable to me.

mkArtakMSFT:
@pranavkm, @javiercn can you guys please review this? Thanks!

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
```diff
@@ -125,40 +125,32 @@ public virtual void EvaluatePreflightRequest(HttpContext context, CorsPolicy pol
...
+                            _logger?.PolicyFailure();
+                            _logger?.RequestHeaderNotAllowed(requestHeader);
+                            return;
+                        }
                     }
                 }
+
+                result.FilterSimpleMethods
+                    = !(CorsConstants.SimpleMethods.Contains(accessControlRequestMethod.ToString(), StringComparer.OrdinalIgnoreCase) &&
+                        requestHeaders.Contains("content-type", StringComparer.OrdinalIgnoreCase));
```
javiercn: Shouldn't we check here for the content-types we know shouldn't trigger a preflight?
```
application/x-www-form-urlencoded
multipart/form-data
text/plain
```

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
```diff
@@ -201,17 +193,25 @@ public virtual void ApplyResult(CorsResult result, HttpResponse response)
 
             if (result.AllowedMethods.Count > 0)
             {
-                // Filter out simple methods
-                var nonSimpleAllowMethods = result.AllowedMethods
-                    .Where(m =>
-                        !CorsConstants.SimpleMethods.Contains(m, StringComparer.OrdinalIgnoreCase))
-                    .ToArray();
+                string[] allowMethods;
+                if (result.FilterSimpleMethods)
```
javiercn: Could this be done in https://github.com/aspnet/CORS/pull/174/files#diff-a347a5f04243b17ec20cf0896f393f68R159 directly? That way we would save the public API bit and the fix would be more straight forward

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
```diff
@@ -125,40 +125,32 @@ public virtual void EvaluatePreflightRequest(HttpContext context, CorsPolicy pol
...
-
-                if (!found)
-                {
-                    _logger?.PolicyFailure();
-                    _logger?.AccessControlMethodNotAllowed(accessControlRequestMethod);
-                    return;
-                }
+                _logger?.PolicyFailure();
+                _logger?.AccessControlMethodNotAllowed(accessControlRequestMethod);
+                return;
```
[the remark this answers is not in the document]
FlyingWraptor: A for loop seemed to be an useless workaround used for filtering using ignorecase. I just changed it to whatever the rest was using as a reflex, didn't know I commited this for this fix. I'm happy to reverting it if you think it needs reverting.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
FlyingWraptor: Preflight requests do not send this information.
As per spec the client is the one enforcing this check, if it doesn't use one of those 3 mime-types it'll do a request using the Access-Control-Allow-Headers: content-type header, not specifying the content-type because it will do that in the actual request.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
FlyingWraptor: AddResult gets called after the EvaluatePreflight removing these cors 'Simple Requests'. So unless a rewrite doing this check in both the EvaluateRequest aswell as the EvaluatePreflightRequest.

As far as I know there's no need for the filtering of Simple Methods outside of Preflight requests (But I'm not sure, which is why I left it in the general AddResult() function using this patch.

pranavkm: We were discussing possibly not doing the filtering at all. Like you said, the spec doesn't require it. I have some reservations about doing this by default (with no "flag" to opt out 2.1). Let me get back to you by early next week once we've had a chance to discuss this.

FlyingWraptor: I was thinking about that too, Filtering is just adding extra overhead and the client is supposed to de termine if it needs to do a preflight request or not. I'll be waiting for a reply and I'm happy to remove these filters for and push them to this issue.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
javiercn: @FlyingWraptor Thanks for the clarification. You are totally right.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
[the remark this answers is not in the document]
javiercn: @FlyingWraptor We just don't want to change code that is not related to a given fix as it can have unexpected consequences that require further testing. (No matter how small or trivial it seems).

I hope that makes sense.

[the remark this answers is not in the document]
FlyingWraptor: Understood, will revert right now.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
pranavkm: We had a chat and landed on removing the filtering in entirety. The change here should be pretty-trivial - if you start afresh from release/2.2:

* Remove the filtering for methods, headers, and exposed headers
* Remove `CorsConstants.Simple*`

Does this sound like something you'd like to continue working on?

FlyingWraptor:
@ pranavkm Will these changes suffice? An extra test wasn't needed, actually some of them required removal. But I made one validating that the filter indeed is non-existant anyways.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
```diff
@@ -151,8 +151,7 @@ public virtual void EvaluatePreflightRequest(HttpContext context, CorsPolicy pol
             {
                 foreach (var requestHeader in requestHeaders)
                 {
-                    if (!CorsConstants.SimpleRequestHeaders.Contains(requestHeader, StringComparer.OrdinalIgnoreCase) &&
```
pranavkm: I hadn't noticed `SimpleRequestHeaders` being used as an additional allow list. @Eilon \ @javiercn should we leave the existing behavior since it's breaking? I'm all for removing magical behavior (and this is entirely magical since it's second guessing the user), but seems like it's less cost for us to keep it for now.

The specific scenario that's affected by this is when `AllowAnyHeader = false`, but the configured allow headers does not include one of the following - `Origin`, `Accept`, `Accept-Language`, or `Content-Language`. CORS will start rejecting these requests where it previously did.

FlyingWraptor: @pranavkm  How's that if I may ask? line 154 checks the configured headers.
Also you might be talking about the filterin from (removed) lines 220 to 231 which made that rejecting occur or am I wrong?

FlyingWraptor: @pranavkm I agree, then this line should stay in here as this is kinda spec behavior isn't it?

Eilon: I agree with removing magic. But we should post an announcement for this.

FlyingWraptor: So what is the play @pranavkm? Reverting the change or keeping it like this?

pranavkm: Keeping it as is seems is fine. I'll get this merged in as soon as I get a change to draft an announcement.

FlyingWraptor: Does CORS reject requests if it doesn't contain any of the 'Simple headers' or if it doesn't contain all the 'Simple headers'?

pranavkm:
Thanks for the PR @FlyingWraptor!

PR #177, opened 2018-08-18
Clarify CORS policy result log messages

cmeeren:
Fixes https://github.com/aspnet/Home/issues/3419

@natemcmaster


natemcmaster:
It looks like the tests need to be updated to match.

PR #179, opened 2018-09-25
Remove filtering from CorsService

pranavkm:
* Add browser based integration tests


samples/SampleDestination/Startup.cs
```diff
@@ -5,34 +5,74 @@
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Logging;
 
 namespace SampleDestination
 {
     public class Startup
     {
+        private const string DefaultAllowedOrigin = "http://localhost:9001";
```
javiercn: I do believe there are issues using localhost for CORS, but I don't remember the details. @rynowak can probably comment on this. Given that you are using Kestrel. Could you use the machine name for this?

javiercn: If you listen on http://+:9001 you should be able to just to http://machine-name:9001 I believe

javiercn: I tried with the machine name and works fine, so I think it's better to use the machine name just in case.

samples/SampleOrigin/Startup.cs
```diff
@@ -17,13 +17,8 @@ public void ConfigureServices(IServiceCollection services)
...
         public void Configure(IApplicationBuilder app, IHostingEnvironment env)
         {
-            app.Run(context =>
-            {
-                var fileInfoProvider = env.WebRootFileProvider;
-                var fileInfo = fileInfoProvider.GetFileInfo("/Index.html");
-                context.Response.ContentType = "text/html";
-                return context.Response.SendFileAsync(fileInfo);
-            });
+            app.UseDefaultFiles();
```
javiercn: Do you need this?

samples/SampleOrigin/wwwroot/Index.html
```diff
@@ -1,89 +0,0 @@
...
-    </style>
-    <title>CORS Sample</title>
-</head>
-<body>
-    <script type="text/javascript">
-        // Make the CORS request.
-        function makeCORSRequest(method, headerName, headerValue) {
-            // Destination server with CORS enabled.
-            var url = 'http://destination.example.com:5000/';
-            var request = new XMLHttpRequest();
```
javiercn: We should also test the fetch API as I believe there are differences in the behavior in this area

javiercn: Nevermind

samples/SampleOrigin/wwwroot/Test.htm
```diff
@@ -0,0 +1,137 @@
...
+            function _fetch(options) {
+                const url = 'http://localhost:9000/' + options.testPath;
+                delete options.testPath;
+                options.cors = 'cors';
+
+                return fetch(url, options);
+            }
+
+            function assertSuccess(options) {
+                return _fetch(options).then(s => true).catch(s => false);
```
javiercn: What is the error that you get when the request fails. Is there a way to be more specific?

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
```diff
@@ -26,7 +27,7 @@ public class CorsService : ICorsService
         /// </summary>
         /// <param name="options">The option model representing <see cref="CorsOptions"/>.</param>
         public CorsService(IOptions<CorsOptions> options)
```
javiercn: What's the point of this constructor? Can we obsolete/kill it?

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
```diff
@@ -78,12 +79,23 @@ public CorsResult EvaluatePolicy(HttpContext context, CorsPolicy policy)
                 throw new ArgumentNullException(nameof(policy));
             }
 
-            var corsResult = new CorsResult();
-            var accessControlRequestMethod = context.Request.Headers[CorsConstants.AccessControlRequestMethod];
-            if (string.Equals(context.Request.Method, CorsConstants.PreflightHttpMethod, StringComparison.OrdinalIgnoreCase) &&
-                !StringValues.IsNullOrEmpty(accessControlRequestMethod))
+            var origin = context.Request.Headers[CorsConstants.Origin];
+            // A CORS-preflight request ... uses `OPTIONS` as method and includes these headers:
```
javiercn: Not sure what you mean with the ellipsis

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
```diff
@@ -78,12 +79,23 @@ public CorsResult EvaluatePolicy(HttpContext context, CorsPolicy policy)
...
+            // Access-Control-Request-Method, Access-Control-Request-Headers
+            // To keep parity with legacy behavior, we'll ignore the presence of Access-Control-Request-Headers header. 
+            var requestHeaders = context.Request.Headers;
+            var isPreflightRequest =
+                string.Equals(context.Request.Method, CorsConstants.PreflightHttpMethod, StringComparison.OrdinalIgnoreCase) &&
+                requestHeaders.ContainsKey(CorsConstants.AccessControlRequestMethod);
```
javiercn: Why did we switch from checking IsNullOrEmpty to just ContainsKey? I imagine that `Access-Control-Request-Method:` will fail later down the pipe, but just checking.

I do believe that Access-Control-Request-Headers is not mandatory, it's probably why it's not being check there.

I think the overall comment above might be misleading (or at least I don't understand it). It seems to imply that we are changing the behavior, but I don't see how the behavior is different.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
```diff
@@ -78,12 +79,23 @@ public CorsResult EvaluatePolicy(HttpContext context, CorsPolicy policy)
...
+
+            var corsResult = new CorsResult
+            {
+                IsPreflightRequest = isPreflightRequest,
+                IsCorsResponseAllowed = IsOriginAllowed(policy, origin),
+            };
+
+            if (isPreflightRequest)
             {
-                _logger?.IsPreflightRequest();
```
javiercn: Why are we removing logging from here?

src/Microsoft.AspNetCore.Cors/Internal/CORSLoggerExtensions.cs
```diff
@@ -64,6 +66,11 @@ static CORSLoggerExtensions()
                 LogLevel.Warning,
                 9,
                 "Failed to apply CORS Response headers.");
+
+            _noCorsPolicyFound = LoggerMessage.Define(
+                LogLevel.Information,
+                new EventId(10, nameof(CorsMiddleware)),
```
Eilon: The event name must *also* be unique amongst names, so using `nameof(CorsMiddleware)` is probably not a good idea.

Eilon: (So either remove the event name, or give it a hard-coded name that would be unique within its namespace.)

samples/SampleOrigin/wwwroot/Test.htm
```diff
@@ -0,0 +1,135 @@
...
+                    method: 'DELETE'
+                }),
+
+                "Allowed Header": assertSuccess({
+                    testPath: 'allow-header-method',
+                    method: 'PUT',
+                    headers: new Headers({ "X-Test": "value", 'Content-Type': 'application/json' })
+                }),
+
+                // This one is weird - although the server performs a preflight request and receives a Access-Control-Allow-Methods: PUT,
```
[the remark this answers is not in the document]
pranavkm: I don't know. POST is considered a safe-listed method (https://fetch.spec.whatwg.org/#methods), so maybe the browser ignores it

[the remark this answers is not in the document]
javiercn: POST is considered safe as long as the content type is form url encoded. Other content-types are not considered safe.
Basically, POST is considered safe as long as is something you can write in a plain old HTML form.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
```diff
@@ -78,12 +79,23 @@ public CorsResult EvaluatePolicy(HttpContext context, CorsPolicy policy)
...
+            // To keep parity with legacy behavior, we'll ignore the presence of Access-Control-Request-Headers header. 
+            var requestHeaders = context.Request.Headers;
+            var isPreflightRequest =
+                string.Equals(context.Request.Method, CorsConstants.PreflightHttpMethod, StringComparison.OrdinalIgnoreCase) &&
+                requestHeaders.ContainsKey(CorsConstants.AccessControlRequestMethod);
+
+            var corsResult = new CorsResult
+            {
+                IsPreflightRequest = isPreflightRequest,
+                IsCorsResponseAllowed = IsOriginAllowed(policy, origin),
```
javiercn: This is misleading. The response might not be allowed even if the origin is allowed. Is there a better name for this?

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
```diff
@@ -181,86 +169,69 @@ public virtual void ApplyResult(CorsResult result, HttpResponse response)
...
-            {
-                headers[CorsConstants.AccessControlAllowOrigin] = result.AllowedOrigin;
-            }
-
-            if (result.VaryByOrigin)
+            if (!result.IsCorsResponseAllowed)
             {
-                headers["Vary"] = "Origin";
+                // In case a server does not wish to participate in the CORS protocol, its HTTP response to the 
+                // CORS or CORS-preflight request must not include any of the above headers.
```
javiercn: Where are we logging that the preflight was denied?

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
pranavkm: `IsOriginAllowed`?

javiercn: That's definitely better

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
pranavkm: Happens as part of `IsOriginAllowed` which says the policy failed

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
pranavkm: I haven't seen a browser request without the header, but it really isn't a hill to die on 👍

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
pranavkm: Moved it to `IsOriginAllowed` which is where we log if the policy is failed

src/Microsoft.AspNetCore.Cors/Internal/CORSLoggerExtensions.cs
```diff
@@ -64,6 +66,11 @@ static CORSLoggerExtensions()
                 LogLevel.Warning,
                 9,
                 "Failed to apply CORS Response headers.");
+
+            _noCorsPolicyFound = LoggerMessage.Define(
+                LogLevel.Information,
+                new EventId(10, "NoCorsPolicyFound"),
```
pranavkm: Fixed EventId name

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
pranavkm: > Why did we switch from checking IsNullOrEmpty to just ContainsKey?

It seems enough for us to verify the presence of the header - we don't actually use the header value here. If you feel strongly about it, I can add it back. Removed the comment, I agree it didn't help.

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsService.cs
```diff
@@ -78,12 +85,25 @@ public CorsResult EvaluatePolicy(HttpContext context, CorsPolicy policy)
                 throw new ArgumentNullException(nameof(policy));
             }
 
-            var corsResult = new CorsResult();
-            var accessControlRequestMethod = context.Request.Headers[CorsConstants.AccessControlRequestMethod];
-            if (string.Equals(context.Request.Method, CorsConstants.PreflightHttpMethod, StringComparison.OrdinalIgnoreCase) &&
-                !StringValues.IsNullOrEmpty(accessControlRequestMethod))
+            if (policy.AllowAnyOrigin && policy.SupportsCredentials)
+            {
+                _logger.InsecureConfiguration();
```
pranavkm: Added logging here

src/Microsoft.AspNetCore.Cors/Internal/CORSLoggerExtensions.cs
```diff
@@ -64,6 +66,16 @@ static CORSLoggerExtensions()
...
+                "No CORS policy found for the specified request.");
+
+            _insecureConfiguration = LoggerMessage.Define(
+                LogLevel.Warning,
+                new EventId(11, "CorsInsecureConfiguration"),
+                "CORS policy uses an insecure configuration that allows any origin and supports credentials.");
```
pranavkm: cc @Eilon for text of the log

test/Microsoft.AspNetCore.Cors.Test/CorsMiddlewareTests.cs
```diff
@@ -169,14 +169,92 @@ public async Task PreFlight_MatchesPolicy_SetsResponseHeaders()
...
+                        options.AddPolicy("customPolicy", policy);
+                    });
+                });
+
+            using (var server = new TestServer(hostBuilder))
+            {
+                // Act
+                // Preflight request.
+                var response = await server.CreateRequest("/")
+                    .AddHeader(CorsConstants.Origin, "http://localhost:5001")
```
javiercn: Can you avoid using localhost in these tests altogether? It's confusing because some browsers don't do CORS over localhost for http.

src/Microsoft.AspNetCore.Cors/Internal/CORSLoggerExtensions.cs
Eilon: What exactly is this message trying to say? Is it recommending that the user not do this? If so, it doesn't really seem to say what the effective problem is, nor how to fix it.

pranavkm: > CORS protocol does not allow specifying a wildcard (any) origin and credentials. Configure the policy by listing individual origins if credentials needs to be supported.

Does that make more sense?

Eilon: How about:

> The CORS protocol does not allow specifying a wildcard (any) origin and credentials at the same time. Configure the policy by listing individual origins if credentials needs to be supported.

FlyingWraptor:
Happy to see CORS being reworked, it was hard to work with before.

PR #180, opened 2018-10-02
Normalize internationalized domain names when adding to CORS

pranavkm:
Fixes https://github.com/aspnet/Home/issues/3353

### DIFF
--- a/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyBuilder.cs
+++ b/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyBuilder.cs
@@ -40,10 +40,34 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
         {
             foreach (var origin in origins)
             {
-                _policy.Origins.Add(origin.ToLowerInvariant());
+                var normalizedOrigin = GetNormalizedOrigin(origin);
+                _policy.Origins.Add(normalizedOrigin);
             }
 
             return this;
+        }
+
+        internal static string GetNormalizedOrigin(string origin)
+        {
+            if (Uri.TryCreate(origin, UriKind.Absolute, out var uri) &&
+                (uri.Scheme == Uri.UriSchemeHttp || uri.Scheme == Uri.UriSchemeHttps) &&
+                !string.Equals(uri.IdnHost, uri.Host, StringComparison.Ordinal))
+            {
+                var builder = new UriBuilder(uri.Scheme.ToLowerInvariant(), uri.IdnHost.ToLowerInvariant());
+                if (!uri.IsDefaultPort)
+                {
+                    // Uri does not have a way to differentiate between a port value inferred by default (e.g. Port = 80 for http://www.example.com) and
+                    // a default port value that is specified (e.g. Port = 80 for http://www.example.com:80). Although the HTTP or FETCH spec does not say 
+                    // anything about including the default port as part of the Origin header, at the time of writing, browsers drop "default" port when navigating
+                    // and when sending the Origin header. All this goes to say, it appears OK to drop an explicitly specified port, 
+                    // if it is the default port when working with an IDN host.
+                    builder.Port = uri.Port;
+                }
+
+                return builder.Uri.GetComponents(UriComponents.SchemeAndServer, UriFormat.Unescaped);
+            }
+
+            return origin.ToLowerInvariant();
         }
 
         /// <summary>
--- a/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyBuilder.cs
+++ b/src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyBuilder.cs
@@ -17,6 +17,7 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
         /// Creates a new instance of the <see cref="CorsPolicyBuilder"/>.
         /// </summary>
         /// <param name="origins">list of origins which can be added.</param>
+        /// <remarks> <see cref="WithOrigins(string[])"/> for details on normalizing the origin header</remarks>
         public CorsPolicyBuilder(params string[] origins)
         {
             WithOrigins(origins);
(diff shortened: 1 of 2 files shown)


test/Microsoft.AspNetCore.Cors.Test/CorsPolicyBuilderTests.cs
```diff
@@ -299,5 +299,98 @@ public void DisallowCredential_SetsSupportsCredentials_ToFalse()
             var corsPolicy = builder.Build();
             Assert.False(corsPolicy.SupportsCredentials);
         }
+
+        [Fact]
+        public void GetNormalizedOrigin_ReturnsLowerCasedValue_IfStringIsNotParseableAsUrl()
```
Tratcher: If the string is not parsable as a URI then the casing is the least of their worries.

test/Microsoft.AspNetCore.Cors.Test/CorsPolicyBuilderTests.cs
```diff
@@ -299,5 +299,98 @@ public void DisallowCredential_SetsSupportsCredentials_ToFalse()
...
+
+            // Assert
+            Assert.Equal("x:\\test", normalizedOrigin);
+        }
+
+        [Fact]
+        public void GetNormalizedOrigin_ReturnsLowerCasedValue_IsStringIsNetworkPath()
+        {
+            // Arrange
+            var origin = "File://MNT/Share/";
```
Tratcher: How is this a relevant test case?

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyBuilder.cs
```diff
@@ -40,12 +40,36 @@ public CorsPolicyBuilder WithOrigins(params string[] origins)
...
+                !string.Equals(uri.IdnHost, uri.Host, StringComparison.Ordinal))
+            {
+                var builder = new UriBuilder(uri.Scheme.ToLowerInvariant(), uri.IdnHost.ToLowerInvariant());
+                if (!uri.IsDefaultPort)
+                {
+                    // Uri does not have a way to differentiate between a port value inferred by default (e.g. Port = 80 for http://www.example.com) and
+                    // a default port value that is specified (e.g. Port = 80 for http://www.example.com:80). Although the HTTP or FETCH spec does not say 
+                    // anything about including the default port as part of the Origin header, at the time of writing, browsers drop "default" port when navigating
+                    // and when sending the Origin header. All this goes to say, it appears OK to drop an explicitly specified port, 
+                    // if it is the default port when working with an IDN host.
```
Tratcher: This comment doesn't match the GetNormalizedOrigin_PreservesPort test results.

test/Microsoft.AspNetCore.Cors.Test/CorsPolicyBuilderTests.cs
pranavkm: I sort of wanted to make sure the behavior is no different (we don't throw for instance) from what it looks like today.

test/Microsoft.AspNetCore.Cors.Test/CorsPolicyBuilderTests.cs
```diff
@@ -299,5 +299,85 @@ public void DisallowCredential_SetsSupportsCredentials_ToFalse()
             var corsPolicy = builder.Build();
             Assert.False(corsPolicy.SupportsCredentials);
         }
+
+        [Theory]
+        [InlineData(@"Some-String", "some-string")]
+        [InlineData(@"x:\\Test", "x:\\test")]
+        [InlineData("FTP://Some-url", "ftp://some-url")]
+        public void GetNormalizedOrigin_ReturnsLowerCasedValue_IfStringIsNotHttpOrHttpsUrl(string origin, string expected)
```
pranavkm: Better?

test/Microsoft.AspNetCore.Cors.Test/CorsPolicyBuilderTests.cs
```diff
@@ -299,5 +299,85 @@ public void DisallowCredential_SetsSupportsCredentials_ToFalse()
...
+            var normalizedOrigin = CorsPolicyBuilder.GetNormalizedOrigin(origin);
+
+            // Assert
+            Assert.Equal("http://www.example.com", normalizedOrigin);
+        }
+
+        [Theory]
+        [InlineData("http://www.Example.com:80", "http://www.example.com:80")]
+        [InlineData("https://www.Example.com:8080", "https://www.example.com:8080")]
+        public void GetNormalizedOrigin_PreservesPort_ForNonIdnHosts(string origin, string expected)
```
pranavkm: Updated test name

Tratcher: Having different default port behavior for IDNs and non-IDNS doesn't sound like it's actually going to work. From your description above it sounds like dropping the default port is what browsers expect.

pranavkm: My concern was that it's a really large browser matrix to test to be 100% certain. It seemed less problematic with Ids since they weren't working to start with. But it'd be more frustrating if we broke it for apps \ browsers that worked correctly.

Tratcher: You need to explain all of this normalization in the doc comments on WithOrigins.

pranavkm: Added docs

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyBuilder.cs
```diff
@@ -36,6 +37,21 @@ public CorsPolicyBuilder(CorsPolicy policy)
         /// </summary>
         /// <param name="origins">The origins that are allowed.</param>
         /// <returns>The current policy builder.</returns>
+        /// <remarks>
+        /// This method normalizes the origin headers prior to adding it to <see cref="CorsPolicy.Origins"/> to match
```
Tratcher: origin headers -> origin values

src/Microsoft.AspNetCore.Cors/Infrastructure/CorsPolicyBuilder.cs
```diff
@@ -36,6 +37,21 @@ public CorsPolicyBuilder(CorsPolicy policy)
...
+        /// the normalization performed by the browser on the value sent in the <c>ORIGIN</c> header.
+        /// <list type="bullet">
+        /// <item>
+        /// If the specified origin has an internationalized domain name (IDN), the punycoded value is used. If the origin
+        /// specifies a default port (e.g. 443 for HTTPS or 80 for HTTP), this will be dropped as part of normalization.
```
Tratcher: Except the default port is only dropped for IDNs

pranavkm: It says so, doesn't it: 

> If the specified origin has an internationalized domain name (IDN)

PR #188, opened 2018-10-11
Log when a request is an OPTIONS request but not a preflight request

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/2375


src/Microsoft.AspNetCore.Cors/Internal/CORSLoggerExtensions.cs
```diff
@@ -76,6 +77,11 @@ static CORSLoggerExtensions()
                 LogLevel.Warning,
                 new EventId(11, "CorsInsecureConfiguration"),
                 "The CORS protocol does not allow specifying a wildcard (any) origin and credentials at the same time. Configure the policy by listing individual origins if credentials needs to be supported.");
+
+            _isNotPreflightRequest = LoggerMessage.Define(
+                LogLevel.Debug,
+                new EventId(12, "OptionsRequestWithoutAccessControlRequestMethodHeader"),
+                "The request uses HTTP OPTIONS method but does not have a Access-Control-Request-Method header. This request will not be treated as a CORS preflight request.");
```
pranavkm: @Eilon does the text look ok?

PR #191, opened 2018-10-16
Increase test timeout to accomodate slow machines

pranavkm:


pranavkm:
Desktop tests on Win7 machine are pretty slow.

PR #202, opened 2018-11-09
Inject ICorsPolicyProvider instance through Invoke

### LINKED ISSUE #105: CorsMiddleware should resolve dependencies dynamically

pranavkm:
This allows for scoped instances of `ICorsPolicyProvider` to be injected in the CORS middleware and prevents turning them into singletons.

Resolves https://github.com/aspnet/AspNetCore/issues/2377


pranavkm:
Updated version of https://github.com/aspnet/CORS/pull/106

pranavkm:
The code's pretty identical in the PR. I think there's plans for better design here but merging this in before the repo gets mondo-repoed.

